Issue #7, opened 2016-12-05
Add development features


MecuSorin:
Add some development features:

- an easy way to enable/disable some tests or test cases by either marking them with **pending**
```fsharp
/// Allows to mark a test as Pending (will be skipped/ignored if no other TestAttribute is present)
/// Is a fast way to exclude some tests from running.  
/// Should be used just during debugging sessions and when a feature is not yet ready for testing (that adds noise to the test results).
[<AttributeUsage(AttributeTargets.Method ||| AttributeTargets.Property ||| AttributeTargets.Field)>]
type PTestsAttribute() = inherit Attribute()
```
 - an easy way to **focus**(check just some tests or test cases and ignore the rest of the test suite) on development task at hand
```fsharp
/// Allows to ignore all tests and test cases that are not market as Focused
/// Is a fast way to include/exclude some tests from running. 
/// Should be used just during debugging sessions.
[<AttributeUsage(AttributeTargets.Method ||| AttributeTargets.Property ||| AttributeTargets.Field)>]
type FTestsAttribute() = inherit Attribute()
```
The statistics should keep track of the disabled tests (marked with Pending or excluded by some Focused tests)

haf:
Sure, but can't it be done as a union case for `Tests` also? After all, we're focusing on tests as values here.

MecuSorin:
Actually I was looking at the code, on how to implement it and the cleanest way in my opinion is to augument all three variants of the union with a state something like:
```
/// The parent state (watching the tests as a tree structure) will influence
/// the child tests state. By following rules, if parent test state is:
///     - Focused will elevate all Run child tests to Focused. 
///              Focused and Pending child tests will not change state(behavior)
///     - Run will not influence the child tests states(behavior).
///     - Pending will elevate all Run and Focused child tests to Pending.
///              Pending child test will not change state(behavior)
type TestState =
  /// The default state of a test that will be evaluated
  | Run
  /// The state of a test that will be ignored by Expecto
  | Pending
  /// The state of a test that will be evaluated
  /// All other test marked with Run or Pending will be ignored
  | Focused

/// Test tree – this is how you compose your tests as values. Since
/// any of these can act as a test, you can pass any of these DU cases
/// into a function that takes a Test.
type Test =
  /// A test case is a function from unit to unit, that can be executed
  /// by Expecto to run the test code.
  | TestCase of code:TestCode * state:TestState
  /// A collection/list of tests.
  | TestList of tests:Test seq * state:TestState
[...]

PR #9, opened 2016-12-07
Added support for focusing and pending tests

MecuSorin:
I'm sorry that this PR is so big, but is the only way it makes sense.

Only thing that is not covered by tests is discovery of tests marked with *FTestsAttribute* - as will mess up the current tests. I tested manually and it's working :) . I was thinking on creating a new project for automate testing this particular issue.

This PR is covering the issue #7

### DIFF
--- a/Expecto.Tests/FocusedTests.fs
+++ b/Expecto.Tests/FocusedTests.fs
@@ -1,0 +1,61 @@
+module Expecto.Focused
+  open Expecto
+  open Expecto.Tests
+  open Expecto.Impl
+
+  let synth = Expecto.Impl.sumTestResults
+  let failing = fun _ -> 1 ==? 2
+  let working = ignore
+
+  let pendingTests =
+    testList "pending tests" [
+      ptestList "ignored" [
+        ptestCase "ignored" failing
+        testCase "normal" failing
+      ]
+      testCase "normal" working
+    ]
+
+  let focusedTests =
+    testList "focused tests" [
+      ptestCase "ignored" failing
+      testCase "normal" failing
+      ftestCase "focused" working
+      ftestList "focused" [
+        ptestCase "ignored" failing
+        testCase "normal" working
+        ftestCase "focused" working
+        testList "normal" [
+          ptestCase "ignored" failing
+          testCase "normal" working
+          ftestCase "focused" working ]
+        ptestList "ignored" [
+          ptestCase "ignored" failing
+          testCase "normal" failing
+          ftestCase "focused" failing]
+        ftestList "focused" [
+          ptestCase "ignored" failing
+          testCase "normal" working
+          ftestCase "focused" working ]
+      ]
+      testList "normal" [
+        ptestCase "ignored" failing
+        testCase "normal" failing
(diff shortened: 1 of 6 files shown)


Expecto.Tests/FocusedTests.fs
```diff
@@ -0,0 +1,77 @@
+module Expecto.Focused
```
haf: Should have equal sign after module name? Or no indentation of the text below (preferable)

Expecto.Tests/Tests.fs
```diff
@@ -184,8 +184,8 @@ let timeouts =
           match testFromType Dummy.thisModuleType.Value with
           | Some (TestList (
                       Seq.Two (
-                          TestLabel("test B", TestList _),
-                          TestLabel("test A", TestList _)))) -> ()
+                          TestLabel("test B", TestList (_, Normal), Normal),
```
haf: Maybe we should consider only making parts of the DU possible to focus on? Perhaps even just the tests, the list and not the label?

Expecto.Tests/FocusedTests.fs
MecuSorin: Fine with me, I was using a namespace before, then removed and forgot about the identation. Is better without *=*

README.md
```diff
@@ -115,6 +115,71 @@ integrationTests // from MyLib.Tests
...
+### Pending tests
+
+You can mark an individual spec or container as Pending. This will prevent the spec (or specs within the list) from running.
+You do this by adding a `p` before *testCase* or *testList* or `P` before *Tests* attribute(when reflection tests discovery is used).
+
+```fsharp
+open Expecto
+
+[<PTests>]
+let skippedTestFromReflectionDiscovery = testCase "skipped" <| fun _ -> Expect.equal (2+2) 4 "2+2"
```
MecuSorin: It's better the way you suggested

Expecto.Tests/Tests.fs
haf: The `TestLabel` may not need it: `let test = TestCase (test, Normal)`? Here https://github.com/MecuSorin/expecto/commit/290dbf3c967dbdce5c2bf3b2870ab3669025f07b#diff-312b950ecb6e71200586b2408ab6e821R41

haf:
Looking good otherwise! :)

Expecto.Tests/Tests.fs
MecuSorin: Removing focus state from *TestLabel* would break the propagation of *FocusState* for a structure like:
```fsharp
ftestList "all should run" [ TestLabel "b" (test "should run as parent is focused" { 1 ==? 2 })]
ftest "will run" { 1 ==? 1 }
```
The workaround will make the code complex and harder to follow (when we need to propagate the state, we need to decompose childs until a child with state is found and then translate (combine) the states.

haf:
LGTM then!

Expecto.Tests/Tests.fs
haf: Roger that.

PR #12, opened 2016-12-08
update simplest irb use in readme

fbehrens:
`run simpleTest` doesn't work :)


haf:
Thanks for that @fbehrens !

PR #18, opened 2016-12-12
Add more assertions

Krzysztof-Cieslak:
1. Adds following assertions:
* `stringEnds`
* `strignHasLength`
* `sequenceStarts`
* `isAscending`
* `isDescending`

2. Fixes bug in string and sequence equals when actual was longer than expected

3. Adds to readme new assertions, and changes in parameters done as part of previous PRs.


haf:
Thanks!

Issue #3, opened 2016-11-16
String diff support


haf:
Make output like this easier to parse for humans, by showing at what position the strings differ:

```
Should have both the fields as values, and the default value.. Actual value was "Processor.%\ Utilisation,unit=% Core\ 1=0.03,Core\ 2=0.06,Core\ 3=0.139,value=1i 1435362189575692182" but had expected it to be "Processor.%\ Utilisation,unit=% Core\ 1=0.03,Core\ 2=0.06,Core\ 3=0.139,value=1i 1435362189575692182".
```

cloudRoutine:
I've always found FParsec's  error messages particularly satisfying
```
Error in Ln: 1 Col: 10
[1] [2] [c]
         ^
Expecting: integer number (32-bit, signed)
```
so maybe something like

```
[Expected value was]                              
"Processor.%\ Utilisation,unit=% Core\ 1=0.03,Core\ 2=0.06,Core\ 3=0.139,value=1i 1435362189575692182"
                                                          ↑                                                    
                                                Error in Col: 59
[Actual value was]                                        ↓
"Processor.%\ Utilisation,unit=% Core\ 1=0.03,Core\ 2=0.06;Core\ 3=0.139,value=1i 1435362189575692182"
```

forki:
![image](https://cloud.githubusercontent.com/assets/57396/21313840/5fb130c6-c5f4-11e6-91f9-fdf3ebd1018e.png)

it should at least report both strings - not only first char

Krzysztof-Cieslak:
I guess we could display both strings, and then report 1st (all?) position on which they differ.

forki:
yes please display at least the actual value. the expected vaue I can find in the code, but the actual value is definetly only in the app

MNie:
Hi,

something like that is ok?
![image](https://cloud.githubusercontent.com/assets/7689103/21433461/2e6d6bb4-c870-11e6-8e64-15968683c785.png)

Krzysztof-Cieslak:
I wouldn't use `[` and `]` since it's not used anywhere else. 

But otherwise, looks good to me.

PR #33, opened 2016-12-22
String diff support #3

MNie:
#3 
Show expected/actual string in SequenceEquals fail message.

image

### DIFF
--- a/Expecto.Tests/Tests.fs
+++ b/Expecto.Tests/Tests.fs
@@ -44,13 +44,16 @@ let tests =
         let format = "Failing - string with different length"
         let fstText = "Test"
         let sndText = "Tes2"
+        let diffString = "   ↑"
         let test () = Expect.equal fstText sndText format
         let msg = sprintf "%s.
-            Expected value was
+            Expected string to equal
             %A
-            Actual value was
+            %s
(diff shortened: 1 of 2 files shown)


Krzysztof-Cieslak:
You should also update string equality tests in Expecto.Tests.... which should be failing, no idea why they are passing 🙄

MNie:
@Krzysztof-Cieslak I updated string equality specs.

haf:
@MNie Thanks for being involved. I was wondering if you could make it so that 'it points to' the position that first has the difference? Like so:

```
Expected strings to equal:
"119024851"
      ↑                                                    
The strings differ at index 5.
      ↓
"119026851"
```

Possibly even with nice colourisation of the string contents inside the `"` mark and the `5` position value?

MNie:
@haf 
I saw that I edited not an _equal_ method but _sequenceequal_.. :(
So I edited right now a proper method (_equal_), but I also keep changes in _sequenceEqual_ since they could be helpful.
Here is the output for string comparison using equal:
image
Here is the output for string/array comparison using sequenceEqual:
image

haf:
Fixes #3 

Thank you very much @MNie !

Issue #2, opened 2016-11-08
Support async tests


haf:
Schedule all Job/Alt/Async/Task thingies and await their completions from e.g. `Expecto.Hopac`. This would probably require a 2-stage process with the first discovering the test and the second doing an execute-and-fan-in/join operation.

 1. Support non-monadic parallelism
 2. Support async/task/job etc
 3. Support limiting concurrency via configuration flag.

haf:
Stub issue for #26

AnthonyLloyd:
Willing to help. Do you have more idea of how you want this to work in terms of API.

haf:
The tests are run either by enumerating a parallel or a sequenced sequence, which comes down to the `map` function passed in to the anonymous functions that `evalTestList` returns.

I created a scaffolding for you to work in; and refactored a bit of the control flow to make it easier for you to implement the basic lock feature.

We'll have to further consider the plugin infrastructure for monads, especially if we're talking things that can interfere with each other.

Please see the output and individual commits of `git show f0e97c5..0c243c9`, and see what you can make of it.

PR #34, opened 2016-12-24
add Expect.isFasterThan

AnthonyLloyd:


haf:
Thank you! What a wonderful present! 🎄 🎁

haf:
I'm thinking you can rebase this on master and then do #2's first point (the stuff I've added comments about in the code). That way master is always releasable. Then when we can be sure that tests don't interfere by sequencing those that are tagged to be using this expect function, or those running benchmarks (by means of tagging them somehow) we'll merge this PR to master. Does it sound like a plan?

AnthonyLloyd:
Sure. Will take a look over the next few days.

haf:
Would be good if you debase this on master. Otherwise the merge commits misrepresent authorship and history

haf:
Rebase not debase

AnthonyLloyd:
Sorry got into a pickle. Should be resolved now.

AnthonyLloyd:
I did this a non locking way in the end. Hope its ok.

I think it is better to extend it going forward e.g. if you knew the last run times you could order the parallel tests so that the longer ones start first.

haf:
Hi there... This is starting to look ready to merge, right? I didn't notice your new commit; it's a good idea to ping me when you push because I don't watch every commit, but instead need @haf mentions.

AnthonyLloyd:
@haf yes, all good to go and rebased.

There looks to be a dependency problem that has come up from BenchmarkDotNet but I guess you are aware of this already?

haf:
No I haven't noticed the problem; probably because mono is more liberal with assembly bindings. What is it?

Also, we need some docs for this. Finally, I'm going to review the code for usages of printfn or similar and perhaps ask you to fix those. I hope that's OK?

AnthonyLloyd:
Below is the error. I can't see an issue for it open on BenchmarkDotNet.

Happy to put some docs together. As a PR to the README.md?

Any changes you'd like let me know. Relatively light on projects at the mo.

```
Exception: System.IO.FileNotFoundException: Could not load file or assembly or one of its dependencies.
File name: 'System.Threading.Tasks.Extensions, Version=4.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'
  at BenchmarkDotNet.Code.CodeGenerator.Generate (BenchmarkDotNet.Running.Benchmark benchmark) [0x00006] in <e49f316815ef414a8469988c242876f6>:0 
  at BenchmarkDotNet.Toolchains.GeneratorBase.GenerateCode (BenchmarkDotNet.Running.Benchmark benchmark, BenchmarkDotNet.Toolchains.ArtifactsPaths artifactsPaths) [0x00006] in <e49f316815ef414a8469988c242876f6>:0 
  at BenchmarkDotNet.Toolchains.GeneratorBase.GenerateProject (BenchmarkDotNet.Running.Benchmark benchmark, BenchmarkDotNet.Loggers.ILogger logger, System.String rootArtifactsFolderPath, BenchmarkDotNet.Configs.IConfig config, BenchmarkDotNet.Characteristics.IResolver resolver) [0x0001b] in <e49f316815ef414a8469988c242876f6>:0 
```

haf:
Fixed. I've also done a bit of cleanups. Please see the last commit message.

I've released v2.4.0 with your new feature, currently writing release notes for it. Could you write a paragraph for me to include with the release notes on how Student's T-test works for performance testing and what the null hypothesis is when it comes to testing functions, please?

AnthonyLloyd:
Cool. Sure working on it now...

haf:
Also, specifically if you could write a few lines about the different outcomes and the Welch statistic, that would be suave ;)

screen shot 2017-01-03 at 13 11 29

Release notes here: https://github.com/haf/expecto/releases/tag/v2.4.0

Docs here: https://github.com/haf/expecto#performance-module

AnthonyLloyd:
@haf Something like this:

This function makes use of a statistical test called [Welch's t-test](https://en.wikipedia.org/wiki/Welch's_t-test).
It starts with the null hypothesis that the functions mean execution times are the same.
The functions are run alternately increasing the sample size to test this hypothesis.
Once a confidence level of 99.99% is reached that this hypothesis is incorrect it stops and reports the results.
99.99% is chosen such that if a test set has 100 performance tests a false test failure would be reported once in many more than 100 runs.
If the performance is very close the test will declare them equal when there is 99.99% confidence they differ by less than 0.5%.
This results in a performance test that is very quick to run (the greater the difference the quicker it will run).
Also, because it is a relative test it can normally be run across all configurations as part of normal unit testing.

haf:
That's great! Do you think you could interleave/add this description to what I've already written in the Performance-testing readme section, please?

PR #35, opened 2016-12-25
Add containsAll method to expecto

MNie:
I saw that in Expect are available following methods:
* _contain_
* _sequenceEqual_

But the first one looks if sequence contains specific element. Second one checks if sequence is equal to another sequence (position in a list is taken into account)
So maybe it is a good idea to add containsAll method which would look if given sequence contains all elements from a second sequence (not more, not least) and not looking for their position in sequence?
So following tests should return "pass" result:
image
If test fails following errors will be shown:
image

### DIFF
--- a/Expecto.Tests/Tests.fs
+++ b/Expecto.Tests/Tests.fs
@@ -431,6 +431,52 @@ let timeouts =
           assertTestFails (test, Normal)
       ]
 
+      testList "contains all comparison" [
+        test "sequence contains all" {
+          Expect.containsAll [|21;37|] [|21;37|] "test sequence"
+        }
+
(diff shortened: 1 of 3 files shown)


haf:
Yes, that's a good feature 👍 

Could you please add xmldoc in the code and statements to the README as well?

MNie:
@haf  Here You are :)

haf:
Will release after short walk.

haf:
Hmm, I'm just reading through this code and it would seem that you've switched around `actual` and `expected`.

MNie:
Yup, but as I saw You fix that?

haf:
Yes it's fixed

PR #38, opened 2016-12-26
Small fixes around contains all

MNie:
As I saw when we pass data like that:
image

Expecto right now will return:
image

Expected result should be like that:
image
@haf I saw that You refactor a little bit containsAll method so could You take a look at this fix?

### DIFF
--- a/Expecto.Tests/Tests.fs
+++ b/Expecto.Tests/Tests.fs
@@ -443,18 +443,18 @@ let expecto =
         testCase "sequence contains everything expected" <| fun _ ->
           let format = "Sequence should contain one and five"
           let test () =
-            Expect.containsAll [|2; 1; 3|] [| 1; 5 |] format
+            Expect.containsAll [|2; 1; 1|] [| 1; 5 |] format
           let msg =
             sprintf "%s.
     Sequence `actual` does not contain all `expected` elements.
         All elements in `actual`:
-        {1, 2, 3}
+        {1, 1, 2}
         All elements in `expected`:
         {1, 5}
         Missing elements from `actual`:
         {5}
         Extra elements in `actual`:
-        {2, 3}"
+        {1, 2}"
               format
           assertTestFailsWithMsg msg (test, Normal)
       ]
--- a/Expecto.Tests/Tests.fs
+++ b/Expecto.Tests/Tests.fs
@@ -452,9 +452,9 @@ let expecto =
         All elements in `expected`:
         {1, 5}
         Missing elements from `actual`:
-        {5}
+        {1, 2}
         Extra elements in `actual`:
-        {1, 2}"
+        {5}"
               format
           assertTestFailsWithMsg msg (test, Normal)
       ]
--- a/Expecto/Expect.fs
+++ b/Expecto/Expect.fs
@@ -204,7 +204,7 @@ let inline private formatSet<'a> (ss : 'a seq) : string =
     |> String.concat ", "
     |> sprintf "{%s}"
 
-let inline private except(sequence: 'a seq, exceptSeq: 'a seq): 'a list =
+let inline private except(baseSeq: 'a seq, exceptSeq: 'a seq): 'a list =
   let groupByOccurances sequence =
     sequence
     |> Seq.groupBy id
@@ -223,7 +223,7 @@ let inline private except(sequence: 'a seq, exceptSeq: 'a seq): 'a list =
       differNrOfElement element
     else Array.create (snd element) (fst element) |> Array.toList
   
-  sequence
+  baseSeq
   |> groupByOccurances
   |> Seq.map elementsWhichDiffer
   |> List.concat
@@ -235,10 +235,10 @@ let containsAll (actual : _ seq)
                 (expected : _ seq)
                 format =
   let extra, missing =
-    except(expected, actual),
-    except(actual, expected)
-
-  if List.isEmpty missing then () else
+    except(actual, expected),
+    except(expected, actual)
+  let isCorrect = List.isEmpty missing && List.isEmpty extra
+  if isCorrect then () else
 
   sprintf "%s.
     Sequence `actual` does not contain all `expected` elements.
--- a/Expecto/Expect.fs
+++ b/Expecto/Expect.fs
@@ -204,17 +204,39 @@ let inline private formatSet<'a> (ss : 'a seq) : string =
     |> String.concat ", "
     |> sprintf "{%s}"
 
+let inline private except(sequence: 'a seq, exceptSeq: 'a seq): 'a list =
+  let groupByOccurances sequence =
+    sequence
+    |> Seq.groupBy id
+    |> Seq.map (fun (letter, occurances) -> (letter, occurances |> Seq.length))
+  let groupedSndSeq = groupByOccurances exceptSeq
+  
+  let differNrOfElement element =
+    let nrOfElements = (snd element) - (groupedSndSeq |> Seq.find(fun y -> fst y = fst element) |> snd)
+    if nrOfElements > 0 then Array.create nrOfElements (fst element) |> Array.toList
+    else []
+  
+  let elementsWhichDiffer element= 
+    if groupedSndSeq |> Seq.exists ((=) element) then
+      []
+    elif groupedSndSeq |> Seq.exists (fun y -> fst y = fst element) then
+      differNrOfElement element
+    else Array.create (snd element) (fst element) |> Array.toList
+  
+  sequence
+  |> groupByOccurances
+  |> Seq.map elementsWhichDiffer
+  |> List.concat
+
 /// Expects the `actual` sequence to contain all elements from `expected`
 /// sequence (not taking into account an order of elements). Calling this
 /// function will enumerate both sequences; they have to be finite.
 let containsAll (actual : _ seq)
                 (expected : _ seq)
                 format =
-  let axs, exs = List.ofSeq actual, List.ofSeq expected
   let extra, missing =
-    let ixs = axs |> List.filter (fun a -> exs |> List.exists ((=) a))
-    axs |> List.filter (fun a -> not (ixs |> List.exists ((=) a))),
-    exs |> List.filter (fun e -> not (ixs |> List.exists ((=) e)))
+    except(expected, actual),
+    except(actual, expected)
 
   if List.isEmpty missing then () else
 
@@ -228,7 +250,7 @@ let containsAll (actual : _ seq)
         %s
         Extra elements in `actual`:
         %s"
-              format (formatSet axs) (formatSet exs) (formatSet missing) (formatSet extra)
+              format (formatSet actual) (formatSet expected) (formatSet missing) (formatSet extra)
   |> Tests.failtest
 
 /// Expects the `actual` sequence to equal the `expected` one.


haf:
I don't agree with the changes. It seems counterintuitive that containsAll would also assert on the count of particular elements in the set. 

I think a function like you describe could be about the distribution, no?

MNie:
@haf  so You proposed that implementation of containsAll should stays as it is right now and this "fix" to containsAll should be in fact a new function in Expect? For example Expect.distributed?
For me it sounds reasonable.

haf:
Yes, like that... But we're testing a [distribution](https://en.wikipedia.org/wiki/Distribution_(mathematics)#Basic_idea) right? Something like (just sketching the idea out here):

```fsharp
Expect.distribution [| "a"; "a"; "b"; "c"; "c"; "c"; "d" |] <|
  Map [
    "a", 2
    "b", 1
    "d", 1
  ]
  "Should pass because a, b and d are all present and at their right counts"
```

What do you think?

MNie:
@haf I agree with resolution which You proposed. It seems to be more "clear".
But what about message if something went wrong?

for 
```
Expect.distribution [|2; 2; 3|] <| 
  Map [
    2, 3
  ]
```

```
{2, ...}
Extra elements in `actual`:
{3, ...}
```

or better
```
Missing elements in `actual` (element: occurances (missing)):
{ ( 2 : 1 ),
.. }
Extra elements in `actual` (element: occurances (extra)):
{ ( 3 : 1 ),
...}
```
What option do You think is better, or maybe something totally different?

haf:
Let me think it over... I'm thinking we could display a histogram.

MNie:
@haf histogram for example from FSharp.Charting? Sounds fancy, but it wouldn't be annoying to user if he has n distribution specs, all fails and he has n windows with histograms?

haf:
I'm thinking something like ASCII-art. Overlapping would be `"a" | ####---` if we plot it side-ways, for example. If buffer width <= expected count for all expectations then plot discretely/exactly, otherwise plot the ratio. A valid bar would look like `"a" | #######`.

MNie:
What if {expected count - actual count} > buffer width? It should "draws" another line of characters? I'm wonder if it would be more clearly to user than simple write `"a" | 4`. 
It looks like an art for art's sake (option with histogram).

haf:
No, then it prints `#` to the ratio of actual<sub>"a"</sub>/expected<sub>"a"</sub>.

But printing `4 of 6` if there are four found out of six expected works too ;)

MNie:
@haf  after some fixes looks like this right now:
image

The question is, if we want to show info only about "errors" ->
* more occurances than expected
* less occurances than expected

and how many they are? 
As it is in above screenshot,  or give an info that for example `4 of 6 occurances of '6' found`?

haf:
The logic seems right, but I find the output hard to interpret. Do you think you could do something like what I've suggested?

MNie:
@haf something like that
image

or something like that:

image

Where:
* `-` is a single element missing 
* `#` - single occurance of element 
* `/` - single occurance of an element which should not be in `actual`. 

But in my opinion it will looks bad if there are not so many expected occurances. Maybe if we want to have more information about `coverage`, a percent info would be better? 

For example
`for '2' found 3 of expected 6, this is a 50% of expected coverage`
What do You think?

haf:
I'm not very much into the list/array syntax of listing tuples of `value * frequency`, because they're easily confused with values. Instead we should use a syntax that doesn't conflict with F#.

How about just:

```
"a" (2/2) | ##
"b" (1/5) | #----
"d" (1/1) | #
```

?

MNie:
@haf You proposition `(1/4)` is shorter and gives exactly the same information as `found 1 of expected 4`, so I agree. This is how it looks after fixes:
image

Expecto.Tests/Tests.fs
```diff
@@ -459,6 +459,85 @@ let expecto =
           assertTestFailsWithMsg msg (test, Normal)
       ]
 
+      testList "#distributed" [
+        testCase "identical sequence" <| fun _ ->
+          Expect.distributed [|21;37|] <| Map [
```
haf: After some consideration: how about just `Expect.distribution`?

Expecto.Tests/Tests.fs
```diff
@@ -459,6 +459,85 @@ let expecto =
...
+                  (1, 1)
+                  (4, 2)
+                ] <| format
+          let msg =
+            sprintf "%s.
+    Sequence `actual` does not contain all `expected` elements.
+        All elements in `actual`:
+        {2, 2, 4}
+        All elements in `expected` ['item', 'number of expected occurances']:
+        {[1, 1], [2, 1], [4, 2]}
```
haf: How about replacing `{[1, 1], [2, 1], [4, 2]}` with output `{1: 1, 2: 1, 4: 2}` just to remove a bit of cruft?

Expecto.Tests/Tests.fs
```diff
@@ -459,6 +459,85 @@ let expecto =
...
+            Expect.distributed [|2; 2; 4|] <| Map[
+              (2, 1)
+              (4, 1)
+            ] <| format
+          let msg =
+            sprintf "%s.
+    Sequence `actual` does not contain all `expected` elements.
+        All elements in `actual`:
+        {2, 2, 4}
+        All elements in `expected` ['item', 'number of expected occurances']:
```
haf: Is spelled "occurrences"

Expecto.Tests/Tests.fs
```diff
@@ -459,6 +459,85 @@ let expecto =
...
+        testCase "sequence does not contains everything expected" <| fun _ ->
+          let format = "Sequence should contain two and two"
+          let test () =
+            Expect.distributed [|2; 2; 4|] <| Map[
+              (2, 1)
+              (4, 1)
+            ] <| format
+          let msg =
+            sprintf "%s.
+    Sequence `actual` does not contain all `expected` elements.
```
haf: "all" -> "every"

Expecto.Tests/Tests.fs
```diff
@@ -459,6 +459,85 @@ let expecto =
...
+        All elements in `actual`:
+        {2, 2, 4}
+        All elements in `expected` ['item', 'number of expected occurances']:
+        {[2, 1], [4, 1]}
+        Missing elements from `actual`:
+        
+        Extra elements in `actual`:
+        '2' (2/1)\t| #/"
```
haf: I think we can drop the `#/` and #-` ligatures when we have the distribution printed.

Expecto/Expect.fs
```diff
@@ -228,7 +229,77 @@ let containsAll (actual : _ seq)
...
+let inline private except(baseMap: Map<_,int>, exceptMap: Map<_,int>, isExcept: bool) =
+  let getMapValue (map: Map<_, int>) (element: _) = 
+    map.TryFind element |> fun x -> x.Value
+  let getResult found expected = 
+    match isExcept with
+    | true -> (found, expected)
+    | _ -> (expected, found)
+  
+  let differNrOfElement element value =
+    let foundElements = (getMapValue exceptMap element)
```
haf: Unneeded paren

Expecto/Expect.fs
```diff
@@ -228,7 +229,77 @@ let containsAll (actual : _ seq)
         %s
         Extra elements in `actual`:
         %s"
-              format (formatSet axs) (formatSet exs) (formatSet missing) (formatSet extra)
+              format (formatResult axs) (formatResult exs) (formatResult missing) (formatResult extra)
+  |> Tests.failtest
+
+let inline private except(baseMap: Map<_,int>, exceptMap: Map<_,int>, isExcept: bool) =
+  let getMapValue (map: Map<_, int>) (element: _) =
```
haf: Can it be negative?

Expecto/Expect.fs
```diff
@@ -228,7 +229,77 @@ let containsAll (actual : _ seq)
         %s
         Extra elements in `actual`:
         %s"
-              format (formatSet axs) (formatSet exs) (formatSet missing) (formatSet extra)
+              format (formatResult axs) (formatResult exs) (formatResult missing) (formatResult extra)
+  |> Tests.failtest
+
+let inline private except(baseMap: Map<_,int>, exceptMap: Map<_,int>, isExcept: bool) =
+  let getMapValue (map: Map<_, int>) (element: _) = 
+    map.TryFind element |> fun x -> x.Value
```
haf: Use `Map.find` instead

Expecto/Expect.fs
```diff
@@ -228,7 +229,77 @@ let containsAll (actual : _ seq)
         %s
         Extra elements in `actual`:
         %s"
-              format (formatSet axs) (formatSet exs) (formatSet missing) (formatSet extra)
+              format (formatResult axs) (formatResult exs) (formatResult missing) (formatResult extra)
+  |> Tests.failtest
+
+let inline private except(baseMap: Map<_,int>, exceptMap: Map<_,int>, isExcept: bool) =
```
haf: The code style should be:

```
let fn (para : type) (para2 : type2) =
```

with:

 - spaces after parameter names
 - spaces after function name
 - preferring curried application instead of tuples

Could we also rename these parameters? What does "exceptMap" mean? Can we make it more semantically appropriate?

Expecto/Expect.fs
haf: Prefer not using `element : _`; instead just write `element` if the type annotation is not used.

Expecto/Expect.fs
```diff
@@ -228,7 +229,77 @@ let containsAll (actual : _ seq)
...
+  let getMapValue (map: Map<_, int>) (element: _) = 
+    map.TryFind element |> fun x -> x.Value
+  let getResult found expected = 
+    match isExcept with
+    | true -> (found, expected)
```
haf: Unneeded parenthesis.

Expecto/Expect.fs
```diff
@@ -228,7 +229,77 @@ let containsAll (actual : _ seq)
...
+    | _ -> (expected, found)
+  
+  let differNrOfElement element value =
```
haf: diff or different? "No" instead of "Nr".

Expecto/Expect.fs
```diff
@@ -228,7 +229,77 @@ let containsAll (actual : _ seq)
...
+    let foundElements = (getMapValue exceptMap element)
+    let missingElements = value - foundElements
+    if missingElements > 0 then getResult foundElements value
```
haf: What do we gain from extracting this to a separate variable? Is it not as easy to read if we inline that variable in the if-expression?

Expecto/Expect.fs
```diff
@@ -228,7 +229,77 @@ let containsAll (actual : _ seq)
...
+    else (0,0)
```
haf: Unneeded parenthesis.

Expecto/Expect.fs
```diff
@@ -228,7 +229,77 @@ let containsAll (actual : _ seq)
...
+  
+  let elementsWhichDiffer element value = 
+    if exceptMap.ContainsKey(element) then
```
haf: Prefer `Map.containsKey` over `:m.ContainsKey`.

Expecto/Expect.fs
```diff
@@ -228,7 +229,77 @@ let containsAll (actual : _ seq)
...
+      differNrOfElement element value
+    else getResult 0 value
+  
+  let printResult tuple =
```
haf: Please destructure the tuple here to make it more readable.

Expecto/Expect.fs
```diff
@@ -228,7 +229,77 @@ let containsAll (actual : _ seq)
...
+      sprintf "(%d/%d)" (fst tuple) (snd tuple)
+    let printWhole found notFound moreThanExpected =
+      sprintf "%s\t| %s%s%s" baseInfo (new String('#', found)) (new String('-', notFound)) (new String('/', moreThanExpected))
+    match fst tuple > snd tuple with
+    | true -> printWhole (snd tuple) 0 (fst tuple - snd tuple)
+    | _ -> printWhole (fst tuple) (snd tuple - fst tuple) 0
+
+  baseMap
+  |> Map.map elementsWhichDiffer
+  |> Map.filter(fun key value -> snd value <> 0)
```
haf: Similarly destructure here on `value`.

Expecto/Expect.fs
```diff
@@ -228,7 +229,77 @@ let containsAll (actual : _ seq)
...
+  |> Map.toList
+  |> List.map(fun elem -> sprintf "'%A' %s" (fst elem) (snd elem |> printResult))
```
haf: Space needed. Avoid `fst` and `snd` if possible.

Expecto/Expect.fs
```diff
@@ -228,7 +229,77 @@ let containsAll (actual : _ seq)
...
+
+/// Expects the `actual` sequence to contain all elements from `expected` map,
```
haf: Docs need to be updated to match our new implementation.

Expecto/Expect.fs
```diff
@@ -228,7 +229,77 @@ let containsAll (actual : _ seq)
...
+/// first element in every tuple from `expected` map means item which should be
+/// presented in `actual` sequence, the second one means an expected number of occurances
+/// of this item sequence (not taking into account an order of elements). 
+/// Calling this function will enumerate both sequences; they have to be finite.
+let distributed (actual : _ seq)
```
haf: Rename like suggestion above?

README.md
```diff
@@ -254,7 +254,8 @@ This module is your main entry-point when asserting.
    then fail with `format` as an error message together with a description
    of `subject` and `length`.
  - `contains : 'a seq -> 'a -> string -> unit` – Expect the sequence to contain the item.
- - `containsAll: 'a seq -> 'a seq -> string -> unit` - Expect the sequence contains all elements from second sequence (not taking into account an order of elements)
+ - `containsAll: 'a seq -> 'a seq -> string -> unit` - Expect the sequence contains all elements from second sequence (not taking into account an order of elements and number of occurances of elements)
+ - `distributed: 'a seq -> Map<'a, int> -> string -> unit` - Expect the sequence contains all elements from map (first element in tuple is an item expected to be in sequence, second one is a number of it's occurances in sequence). Sequence is not taking into account an order of elements.
```
haf: Also needs to be updated. "Expect [that] the ..." Avoid mixing numbers with colloquial speech: "second one". "its", not "it's". Also, is it the "sequence" that takes into account – not the Expect-module function? Also, if something *is* taking something into account, should it not be the Map that specifies the distribution?

Should we consider just taking a `(string * string) seq` instead of a `Map` to avoid the Map c'tor call at the call-site?

haf:
Ok, I've review it. I'm going to log off for today, but I think we are getting somewhere now.

Expecto.Tests/Tests.fs
MNie: So instead of `'2' (2/1) | #/` it should be `'2' (2/1)`?

Expecto/Expect.fs
MNie: Nope, only positive.

Expecto.Tests/Tests.fs
MNie: done

Expecto/Expect.fs
MNie: In my opinion by extracting it to a separate variable code is more clear than `if value - foundElements > 0 then ...` or `match value - foundElements > 0 with..`

README.md
MNie: @hef I fix'd pull request after review (thanks for that). `Should we consider just taking a (string * string) seq instead of a Map to avoid the Map c'tor call at the call-site?` You mean `('a * uint32) seq` not `(string * string) seq` :)?

haf:
Thank you!

MNie:
@haf  You're welcome :)!

PR #47, opened 2017-01-03
Add tests' source file location 

Krzysztof-Cieslak:
Adds ability to print out test location (source file, line of code). Still work in progress. 

Cons:
* As far as I understand there are certain limitations (it's based on Pdb files so won't work in release configs etc)
* It adds a bit of complexity and additional dependency (`Mono.Cecil`)

Pros:
* It will allow much better editor integrations. (For example in VSCode in watch mode marking which tests are passing and which not)

### DIFF
--- a/Expecto.BenchmarkDotNet/Expecto.BenchmarkDotNet.fsproj
+++ b/Expecto.BenchmarkDotNet/Expecto.BenchmarkDotNet.fsproj
@@ -84,15 +84,15 @@
   <Choose>
     <When Condition="$(TargetFrameworkIdentifier) == '.NETFramework' And ($(TargetFrameworkVersion) == 'v4.5' Or $(TargetFrameworkVersion) == 'v4.5.1' Or $(TargetFrameworkVersion) == 'v4.5.2' Or $(TargetFrameworkVersion) == 'v4.5.3' Or $(TargetFrameworkVersion) == 'v4.6' Or $(TargetFrameworkVersion) == 'v4.6.1' Or $(TargetFrameworkVersion) == 'v4.6.2' Or $(TargetFrameworkVersion) == 'v4.6.3')">
       <ItemGroup>
-        <Reference Include="BenchmarkDotNet.Core">
-          <HintPath>..\packages\BenchmarkDotNet.Core\lib\net45\BenchmarkDotNet.Core.dll</HintPath>
-          <Private>True</Private>
-          <Paket>True</Paket>
-        </Reference>
         <Reference Include="System.Management">
           <Paket>True</Paket>
         </Reference>
         <Reference Include="System.Xml">
+          <Paket>True</Paket>
+        </Reference>
+        <Reference Include="BenchmarkDotNet.Core">
+          <HintPath>..\packages\BenchmarkDotNet.Core\lib\net45\BenchmarkDotNet.Core.dll</HintPath>
+          <Private>True</Private>
           <Paket>True</Paket>
         </Reference>
       </ItemGroup>
@@ -200,8 +200,8 @@
     <When Condition="$(TargetFrameworkIdentifier) == '.NETFramework' And ($(TargetFrameworkVersion) == 'v4.6' Or $(TargetFrameworkVersion) == 'v4.6.1' Or $(TargetFrameworkVersion) == 'v4.6.2')">
       <ItemGroup>
         <Reference Include="System.AppContext">
-          <HintPath>..\packages\System.AppContext\lib\net46\System.AppContext.dll</HintPath>
-          <Private>True</Private>
+          <HintPath>..\packages\System.AppContext\ref\net46\System.AppContext.dll</HintPath>
+          <Private>False</Private>
           <Paket>True</Paket>
         </Reference>
       </ItemGroup>
@@ -209,8 +209,8 @@
     <When Condition="$(TargetFrameworkIdentifier) == '.NETFramework' And $(TargetFrameworkVersion) == 'v4.6.3'">
       <ItemGroup>
         <Reference Include="System.AppContext">
-          <HintPath>..\packages\System.AppContext\lib\net463\System.AppContext.dll</HintPath>
-          <Private>True</Private>
+          <HintPath>..\packages\System.AppContext\ref\net463\System.AppContext.dll</HintPath>
+          <Private>False</Private>
           <Paket>True</Paket>
         </Reference>
       </ItemGroup>
@@ -231,8 +231,8 @@
     <When Condition="$(TargetFrameworkIdentifier) == '.NETFramework' And ($(TargetFrameworkVersion) == 'v4.6' Or $(TargetFrameworkVersion) == 'v4.6.1' Or $(TargetFrameworkVersion) == 'v4.6.2' Or $(TargetFrameworkVersion) == 'v4.6.3')">
       <ItemGroup>
         <Reference Include="System.Console">
-          <HintPath>..\packages\System.Console\lib\net46\System.Console.dll</HintPath>
-          <Private>True</Private>
+          <HintPath>..\packages\System.Console\ref\net46\System.Console.dll</HintPath>
+          <Private>False</Private>
           <Paket>True</Paket>
         </Reference>
       </ItemGroup>
@@ -242,8 +242,8 @@
     <When Condition="$(TargetFrameworkIdentifier) == '.NETFramework' And ($(TargetFrameworkVersion) == 'v4.6' Or $(TargetFrameworkVersion) == 'v4.6.1' Or $(TargetFrameworkVersion) == 'v4.6.2' Or $(TargetFrameworkVersion) == 'v4.6.3')">
       <ItemGroup>
         <Reference Include="System.Diagnostics.FileVersionInfo">
-          <HintPath>..\packages\System.Diagnostics.FileVersionInfo\lib\net46\System.Diagnostics.FileVersionInfo.dll</HintPath>
-          <Private>True</Private>
+          <HintPath>..\packages\System.Diagnostics.FileVersionInfo\ref\net46\System.Diagnostics.FileVersionInfo.dll</HintPath>
+          <Private>False</Private>
           <Paket>True</Paket>
         </Reference>
       </ItemGroup>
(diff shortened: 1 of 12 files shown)


haf:
Needs rebase. Tell me when you've transcended WIP-state.

haf:
The last commit seems to attribute e.g. `testSequenced` to yourself. If you're rebasing, consider making a clean rebase instead of committing after the rebase is over.

Krzysztof-Cieslak:
Fixed rebasing

haf:
Is it possible to have this as a separate library that you can just nuget-reference as part of using Expecto? We can make it very easy to reference.

I'm also thinking it may be possible to decorate the `Test` type rather than always extending it structurally...

```fsharp
type SourceTest =
  | SourceTest of test:Test * location:SourceLocation
```

which then can be folded over to re-create the annotated tree?

Krzysztof-Cieslak:
@haf,

> Is it possible to have this as a separate library that you can just nuget-reference as part of using Expecto? We can make it very easy to reference.

I guess it would be possible but if I were to decided I'd prefer to include in here to make it as friction-less for users as possible. Assuming finding location will be enabled by additional flag most users won't see any changes, I guess it would be most useful for editor integrations.

> I'm also thinking it may be possible to decorate the `Test` type rather than always extending it structurally...

Yes, I hope to change it a bit too - I'm not really happy with current design since I need to propagate location everywhere, and I just need it in TestRunResult (assuming we can live without printing location for `--list-tests`  which is using `FlatTest`... or I can handle finding location for `--list-tests` separately \shrug ) so I should be able to inject function finding location (since it's  `Assembly -> TestCode -> SourceLocation`) on `eval` step - similar to Printer. 

As I've mentioned, still WIP, I'm still investigating design space ;)

Krzysztof-Cieslak:
@haf, I changed things a bit - instead of passing location everywhere I try to get it in `eval` step. Please let me know what do you think about this solution.

Expecto/Expecto.fs
```diff
@@ -7,6 +7,13 @@ open System.Linq
 open System.Runtime.CompilerServices
 open System.Reflection
 
+type SourceLocation ={
```
haf: Newline after `=`. camelCase properties.

Expecto/Expecto.fs
```diff
@@ -7,6 +7,13 @@ open System.Linq
...
+  SourcePath: string
+  LineNumber: int
+}
+with static member Empty = {SourcePath = ""; LineNumber = 0}
```
haf: camelCase properties

Expecto/Expecto.fs
```diff
@@ -505,7 +519,7 @@ module Impl =
       else
         None
 
-    fun (printers: TestPrinters) map ->
+    fun (localization : TestCode -> SourceLocation) (printers: TestPrinters) map ->
```
haf: Localisation means translation normally; can we call it 'locate'?

Expecto/Expecto.fs
```diff
@@ -570,16 +590,16 @@ module Impl =
 
   /// Runs a tree of tests, with parameterized printers (progress indicators) and traversal.
   /// Returns list of results.
-  let eval (printer: TestPrinters) map tests =
+  let eval (localization : TestCode -> SourceLocation) (printer: TestPrinters) map tests =
```
haf: Same here

Expecto/Expecto.fs
```diff
@@ -570,16 +590,16 @@ module Impl =
...
     Test.toTestCodeList tests
-    |> evalTestList printer map
+    |> evalTestList localization printer map
 
   /// Evaluates tests sequentially
-  let evalSeq =
-    eval TestPrinters.Default List.map
+  let evalSeq localization =
+    eval localization TestPrinters.Default List.map
```
haf: Should we not parametise the printer too? /cc @AnthonyLloyd talked about it

Expecto/Expecto.fs
```diff
@@ -806,7 +905,8 @@ module Tests =
       filter    = id
       failOnFocusedTests = false
       printer   = TestPrinters.Default
-      verbosity = LogLevel.Info }
+      verbosity = LogLevel.Info
+      localization = (fun _ -> SourceLocation.Empty) }
```
haf: Unneeded paren

haf:
Besides my minor comments; will it work with mdb files, too?

Krzysztof-Cieslak:
Haven't checked it, but according to Mono.Cecil documentation - "Cecil will try to load the debug store support assembly depending on the runtime (pdb on .net, mdb on Mono) and read and write the appropriate symbols." (https://github.com/jbevain/cecil/wiki/Debug-symbols)

Krzysztof-Cieslak:
Fixed all issues from review but printer parametrization (I guess there should be separate PR for it).

I still need to define printer and flag enabling this.

haf:
Can we make the `Empty` static member/property `empty`?

Expecto/Expecto.fs
```diff
@@ -7,6 +7,13 @@ open System.Linq
 open System.Runtime.CompilerServices
 open System.Reflection
 
+type SourceLocation = {
```
haf: Newline after `=`

Expecto/Expecto.fs
```diff
@@ -163,7 +174,7 @@ module Test =
             then name
             else parentName + "/" + name
         loop fullName testList (computeChildFocusState parentState state) sequenced test
-      | TestCase (test, state) -> {name=parentName; test=test; state=computeChildFocusState parentState state; sequenced=sequenced} :: testList
+      | TestCase (test, state) -> {name=parentName; test=test; state=computeChildFocusState parentState state; sequenced=sequenced;} :: testList
```
haf: If you come across code like this, feel free to format it according to DEVGUIDE.

Expecto/Expecto.fs
```diff
@@ -172,7 +183,7 @@ module Test =
   let rec wrap f =
     function
     | TestCase (test, state) -> TestCase ((f test), state)
-    | TestList (testList, state) -> TestList (List.map (wrap f) testList, state)
+    | TestList (testList, state) -> TestList ((List.map (wrap f) testList), state)
```
haf: In this case, a pipe could actually be nice.

haf:
Nice. I'll merge it after rebase and the nigglers are fixed.

haf:
Single test failure:

```
[21:37:25 ERR] expecto/expectations/#distribution/sequence does not contains everything expected failed in 00:00:00.0073172. 

Test failure strings should equal.

          Expected string to equal:

          "Sequence should contain two and two.

    Sequence `actual` does not contain every `expected` elements.

        All elements in `actual`:

        {2, 2, 4}

        All elements in `expected` ['item', 'number of expected occurrences']:

        {2: 1, 4: 1}

        Extra elements in `actual`:

        '2' (2/1)"

                                                                                                                                                                                                                                                                                                                  ↑

          The string differs at index 295.

          "Sequence should contain two and two.

    Sequence `actual` does not contain every `expected` elements.

        All elements in `actual`:

        {2, 2, 4}

        All elements in `expected` ['item', 'number of expected occurrences']:

        {2: 1, 4: 1}

        Extra elements in `actual`:

        '2' (2/1)"

                                                                                                                                                                                                                                                                                                                  ↑

          String does not match at position 295. Expected char: '\010', but got ' '.
```

Expecto.Tests/Tests.fs
```diff
@@ -531,7 +537,7 @@ let expecto =
         All elements in `expected` ['item', 'number of expected occurrences']:
         {2: 1, 4: 1}
         Missing elements from `actual`:
-        
+
```
Krzysztof-Cieslak: @haf, this is a change that makes test fail. Does we really should depend on having empty lines containing spaces in error messages?

haf: No they should be part of the string on a single line. Or a reflex perhaps? Hmm

Krzysztof-Cieslak:
@haf, Fixed distribution text in hacky way. I'm not super happy with it but I believe problem should be handled in separate PR changing distribution error message - I don't think it should display those part of message which are not not showing actual error for the case (so for example don't display `Missing elements from actual:` + empty line if there are no missing elements).

Anyway, green and ready to merge / final review

Expecto.BenchmarkDotNet/Expecto.BenchmarkDotNet.fsproj
```diff
@@ -242,8 +242,8 @@
     <When Condition="$(TargetFrameworkIdentifier) == '.NETFramework' And ($(TargetFrameworkVersion) == 'v4.6' Or $(TargetFrameworkVersion) == 'v4.6.1' Or $(TargetFrameworkVersion) == 'v4.6.2' Or $(TargetFrameworkVersion) == 'v4.6.3')">
       <ItemGroup>
         <Reference Include="System.Diagnostics.FileVersionInfo">
-          <HintPath>..\packages\System.Diagnostics.FileVersionInfo\lib\net46\System.Diagnostics.FileVersionInfo.dll</HintPath>
-          <Private>True</Private>
+          <HintPath>..\packages\System.Diagnostics.FileVersionInfo\ref\net46\System.Diagnostics.FileVersionInfo.dll</HintPath>
```
haf: Cheese moved?

haf:
Thanks Krzysztof!

Krzysztof-Cieslak:
@haf, can we release it? I'd like to release updates to Ionide project using this feature 
error_notifications

Krzysztof-Cieslak:
@AnthonyLloyd running rake build should generate assembly info file

AnthonyLloyd:
ah, didn't know I had to do that

AnthonyLloyd:
sorry, I'm a ruby/rake idiot. Went through a lot of hoops installing bundler via SSL bug manual install.
Now rake build doesn't do much and just updates Gemfile.lock:

C:\Users\Ant\src\expecto [(2b3c161...) +0 ~1 -0 !]> git diff
diff --git a/Gemfile.lock b/Gemfile.lock
index fe9a18f..080c47c 100644
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -10,14 +10,17 @@ GEM
     mini_portile2 (2.1.0)
     nokogiri (1.6.8.1)
       mini_portile2 (~> 2.1.0)
+    nokogiri (1.6.8.1-x64-mingw32)
+      mini_portile2 (~> 2.1.0)
     rake (10.5.0)
     semver2 (3.4.2)

 PLATFORMS
   ruby
+  x64-mingw32

 DEPENDENCIES
   albacore (~> 2.6.2)

 BUNDLED WITH
-   1.13.6
+   1.13.7
C:\Users\Ant\src\expecto [(2b3c161...) +0 ~1 -0 !]> rake build --trace
** Invoke build (first_time, not_needed)

Krzysztof-Cieslak:
@AnthonyLloyd,

I also got those changes in `gemfile.lock`, don't commit them to repo. @haf would now better if they should be committed or not.

Run: `bundle exec rake`

Actually, I think that build description used to be in readme but got removed and not added to DEVGUIDE

AnthonyLloyd:
working thanks

AnthonyLloyd:
I still see [:0] for --summary-location tho

Krzysztof-Cieslak:
All projects here are built with Release configuration. As I've mentioned in first post there are some limitations to this solution since it'is based on Pdb / Mdb files.

PR #52, opened 2017-01-04
Add counts to summary

forki:
[12:26:47 INF] EXPECTO?! Summary...
    Passed:
    	debug-config/sound check - can take screenshot
    	debug-config/search/can go to search
    Ignored:
    	
    Failed:
    	
    Errored:

will be:

    [12:26:47 INF] EXPECTO?! Summary...
    Passed: 2
    	debug-config/sound check - can take screenshot
    	debug-config/search/can go to search
    Ignored: 0
    	
    Failed: 0
    	
    Errored: 0

### DIFF
--- a/Expecto/Expecto.fs
+++ b/Expecto/Expecto.fs
@@ -335,17 +335,21 @@ module Impl =
       duration = results |> Seq.map (fun r -> r.duration) |> Seq.fold (+) TimeSpan.Zero }
 
   let logSummary (summary : TestResultSummary) =
-    let passed = summary.passed |> String.concat "\n\t"
+    let handleLineBreaks elements = 
+        let text = elements |> String.concat "\n\t"
+        if text = "" then text else text + "\n"
+
+    let passed = summary.passed |> handleLineBreaks
     let passedCount = summary.passed |> List.length
-    let ignored = summary.ignored |> String.concat "\n\t"
+    let ignored = summary.ignored |> handleLineBreaks
     let ignoredCount = summary.ignored |> List.length
-    let failed = summary.failed |> String.concat "\n\t"
+    let failed = summary.failed |> handleLineBreaks
     let failedCount = summary.failed |> List.length
-    let errored = summary.errored |> String.concat "\n\t"
+    let errored = summary.errored |> handleLineBreaks
     let erroredCount = summary.errored |> List.length
 
     logger.info (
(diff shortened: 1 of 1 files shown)


Krzysztof-Cieslak:
Any reason for that? Same information is printed out one line above, in standard Expecto results output.

forki:
yes because it looks weird. but maybe it's just linebreaks that needs adjustment

forki:
[12:26:47 INF] EXPECTO?! Summary...
    Passed: 2
    	debug-config/sound check - can take screenshot
    	debug-config/search/can go to search
    Ignored: 0
    Failed: 0	
    Errored: 0

haf:
If we're changing them, then I think they should vertically align. Can you pad the numbers to have their most significant digit at the same column?

forki:
No idea how that goes with the logging framework.

Am 04.01.2017 19:43 schrieb "Henrik Feldt" <notifications@github.com>:

If we're changing them, then I think they should vertically align. Can you
pad the numbers to have their most significant digit at the same column?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/haf/expecto/pull/52#issuecomment-270450818>, or mute
the thread

Krzysztof-Cieslak:
Just add spaces ;]

adamchester:
to use spaces/align right on the numbers you would probably have to make the summary just a string rather than embedded properties in the log message. That makes it less pretty.

I did [add align support to the facade](https://github.com/logary/logary/issues/166#issuecomment-257123187) (among other things) but we decided it was too much [ugly code](https://github.com/logary/logary/compare/master...adamchester:facade-template-event) for the facade. I think adding alignment (separate from the other things I had in that PoC) would still be useful, if @haf agrees.

With that feature align in the facade, you would simply do it like: `{passedCount,-4}`

adamchester:
If you string.Format the number (including spaces) prior to setting the template field, that would work too (but it would be nice if the template field remained a number)

forki:
done

haf:
Thanks!

forki:
could you please hit release button? Would like to test it in my real world tests

haf:
So #53 works for you then?

Krzysztof-Cieslak:
Hey, Travis is with me ;P Maybe @forki is going crazy since it's working only on his machine ;P

forki:
Tbf it's working on multiple machines with 2 different projects.

PR #55, opened 2017-01-05
link to fuchu

forki:


haf:
I don't see the point of this PR. I don't want newcomers to click onwards to the old testing library so a link like this steals attention from the message.

PR #58, opened 2017-01-05
Fix line breaks

forki:
This is the current situation:

image

I hope line breaks are now ok

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -3,8 +3,10 @@
 System.IO.Directory.SetCurrentDirectory __SOURCE_DIRECTORY__
 open System
 open Fake
+open Fake.Testing.Expecto
 
-
+// Pattern specifying assemblies to be tested using expecto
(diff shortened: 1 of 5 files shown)


haf:
Please write a test for this code change.

forki:
will do!

forki:
@haf @Krzysztof-Cieslak I'm pretty close, but how can I go from Message to String? ToString just gives me the type.

forki:
Ok found a way by adding a helper function.
I also enabled FAKE's expecto test helper for dogfooding.

build.fsx
```diff
@@ -53,6 +55,11 @@ Target "Build" (fun _ ->
   |> ignore
 )
 
+Target "RunTests" (fun _ ->
+    !! testExecutables
+    |> Expecto (fun p -> { p with Parallel = false } )
```
haf: I believe Expeco's own tests should be runnable in parallel. :)

PR #60, opened 2017-01-05
fix mutable sync/async logging and support async tests

AnthonyLloyd:
Hopefully the style is ok.

I think for LF vs CRLF you need a .gitattributes file.

I think ultimately there must be a better way to handle sync/async logging. Needs to be able to be switched  from async flush to manual flush.

### DIFF
--- a/Expecto/Expect.fs
+++ b/Expecto/Expect.fs
@@ -434,14 +434,13 @@ let isFasterThanSub (f1:Performance.Measurer<_,_>->'a) (f2:Performance.Measurer<
     Tests.failtestf "%s. Expected f1 (%s) to be faster than f2 (%s) but is ~%.0f%% slower."
                     format (toString s1) (toString s2) ((s1.mean/s2.mean-1.0)*100.0)
   | Performance.MetricLessThan (s1,s2) ->
-    Impl.logger.infoWithBP (
+    Impl.logger.log Info (
       eventX "{message}. f1 ({sample1}) is {percent} faster than f2 ({sample2})."
       >> setField "message" format
       >> setField "sample1" (toString s1)
       >> setField "percent" (sprintf "~%.0f%%" ((1.0-s1.mean/s2.mean)*100.0))
       >> setField "sample2" (toString s2))
     |> Async.StartImmediate
-    //printfn "%s. f1 (%s) is ~%.0f%% faster than f2 (%s)." format (toString s1) ((1.0-s1.mean/s2.mean)*100.0) (toString s2)
 
 /// Expects function `f1` is faster than `f2`. Statistical test to 99.99%
 /// confidence level.
--- a/Expecto/Expect.fs
+++ b/Expecto/Expect.fs
@@ -235,24 +235,24 @@ let containsAll (actual : _ seq)
   |> Tests.failtest
 
 let inline private except (elementsToCheck: Map<_,uint32>) (elementsToContain: Map<_,uint32>) (isExcept: bool) =
-  let getMapValue (map: Map<_, uint32>) (element) = 
+  let getMapValue (map: Map<_, uint32>) (element) =
     map |> Map.find element
-  let getResult found expected = 
+  let getResult found expected =
     match isExcept with
     | true -> found, expected
     | _ -> expected, found
-  
+
   let noOfFoundElements element value =
     let foundElements = (getMapValue elementsToContain element)
     match value > foundElements with
     | true -> getResult foundElements value
     | _ -> 0ul,0ul
-  
-  let elementsWhichDiffer element value = 
+
+  let elementsWhichDiffer element value =
     match elementsToContain |> Map.containsKey(element) with
     | true -> noOfFoundElements element value
     | _ -> getResult 0ul value
-  
+
   let printResult found =
     sprintf "(%d/%d)" (fst found) (snd found)
 
@@ -266,7 +266,7 @@ let inline private except (elementsToCheck: Map<_,uint32>) (elementsToContain: M
 /// first element in every tuple from `expected` map means item which should be
 /// presented in `actual` sequence, the second element means an expected number of occurrences
 /// of this item in sequence.
-/// Function is not taking into account an order of elements. 
+/// Function is not taking into account an order of elements.
 /// Calling this function will enumerate both sequences; they have to be finite.
 let distribution (actual : _ seq)
                 (expected : Map<_,uint32>)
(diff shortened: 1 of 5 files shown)


AnthonyLloyd:
I think I know how this should work. I've just been trying to add support for async tests.

The executions of a test with logging should be made into an async<TestRunResult> which is easy enough.
The parallel ones then can be run together and the sequenced one by one completing their logging.
This should simplify the code and the TestPrinters are now correct with their Async<unit> return values.

Expecto/Expecto.fs
```diff
@@ -544,8 +527,12 @@ module Impl =
 
     fun (printers: TestPrinters) map ->
 
+      let log test =
+        if test.sequenced then Async.StartImmediate else Async.Start
```
haf: It's really just the join at the end of the run that should decide the method with which logs are awaited... Just because we sequence tests, we don't have to start their logs immediately. Also, `logger.infoWithBP` will only back-pressure the logger's buffer, not the printing. So the returned `Async<unit>` from such a function call will yield as soon as the buffer has accepted the message – a final flush of the logging infrastructure at the end should then be enough.

haf:
I really like these changes. However, we may have to think about it a bit; see my in-code comment.

AnthonyLloyd:
Sure. I've been working on the async tests. I'll add that to the PR when its ready and we can discuss.

Currently having some issues with the run taking 20s insead of 2s (logging ACK? not sure)

haf:
The two are identical in the Facade: https://github.com/logary/logary/blob/master/src/Logary.Facade/Facade.fs#L703-L710

AnthonyLloyd:
@haf
I've had to limit the number of threads to get this to not take 20s.
I'm not sure why this is necessary.

Could you take a look at some point and let me know if you think I'm on the right track.

AnthonyLloyd:
@haf,

Just to update you on the status of this.

I've added number of parallel workers to command line and async test tests. Its basically done as far I can see apart from some docs.

There are a few issues outstanding:

- getMethodName - I can't write the code to locate the async tests. --summary-location just gives [:0] for me in Debug and Release for sync tests

- The BenchmarkDotNet tests are still not working. They pass but are excepting and nan in travis.

- There are some tests that are failing for me but not travis (even when I checkout haf/master):

Failed:   7
        all/string comparison/different length, actual is shorter [:0]
        all/string comparison/different length, actual is longer [:0]
        all/string comparison/fail - different content [:0]
        expecto/expectations/#containsAll/sequence contains everything expected [:0]
        expecto/expectations/#distribution/sequence doesn't contain repeats in expected [:0]
        expecto/expectations/#distribution/sequence does contain repeats in expected but should not [:0]
        expecto/expectations/#distribution/sequence does not contains everything expected [:0]

haf:
Your failures may be fixed w/ string changes in #64

haf:
No 64 is merged now.

AnthonyLloyd:
rebased but I still see the errors. May be down to LF vs CRLF:

```
[20:03:12 ERR] expecto/expectations/#distribution/sequence does contain repeats in expected but should not failed in 00:00:00.0855580.
Test failure strings should equal.
          Expected string to equal:
          "Sequence should contain two, two and four.
    Sequence `actual` does not contain every `expected` elements.
        All elements in `actual`:
        {2, 2}
        All elements in `expected` ['item', 'number of expected occurrences']:
        {2: 2, 4: 1}
        '4' (0/1)"

                                                                                                          ↑
          The string differs at index 306.
          "Sequence should contain two, two and four.
    Sequence `actual` does not contain every `expected` elements.
        All elements in `actual`:
        {2, 2}
        All elements in `expected` ['item', 'number of expected occurrences']:
        {2: 2, 4: 1}
        '4' (0/1)
" C:\Users\Ant\src\expecto\Expecto\Expect.fs(301,1): Expecto.Expect.distribution[a](IEnumerable`1 actual, FSharpMap`2 expected, String format)

                                                                                                          ↑
          String `actual` was longer than expected, at pos 306 found item '\010'.
  C:\Users\Ant\src\expecto\Expecto.Tests\Tests.fs(518,1): Expecto.Tests.expecto@518-173.Invoke(String msg)

```

Expecto/Expect.fs
```diff
@@ -434,13 +434,13 @@ let isFasterThanSub (f1:Performance.Measurer<_,_>->'a) (f2:Performance.Measurer<
     Tests.failtestf "%s. Expected f1 (%s) to be faster than f2 (%s) but is ~%.0f%% slower."
                     format (toString s1) (toString s2) ((s1.mean/s2.mean-1.0)*100.0)
   | Performance.MetricLessThan (s1,s2) ->
-    Impl.logger.info (
+    Impl.logger.log Info (
```
haf: If we expect this to be sent all the way to the terminal before the process exists, `logWithAck` should be used.

Expecto/Expecto.fs
```diff
@@ -395,11 +436,12 @@ module Impl =
     createSummaryMessage summary Info
     |> Expecto.Logging.Formatting.defaultFormatter
 
-  let logSummary (log: Log) (summary: TestResultSummary) =
+  let logSummary (summary: TestResultSummary) =
     createSummaryMessage summary
-    |> log.write Info
+    |> logger.log Info
```
haf: LogWithAck

haf:
Do you still have issues with speed?

AnthonyLloyd:
No its fine with the parallel-workers defaulting to logical processors.
I always thought you never needed to do this with Async because of the TPL only creating a fixed number of threads.
All other unit testing libraries have a config for workers and default to no logical processors.

I'll add docs for this command line later and test if a .gitattributes with `git config core.autocrlf false` fixes my local test failure.

haf:
This is a good PR, I've merged it and tested it to some extent and it's looking good. I've also added the `testCaseAsync` test function to the README.

AnthonyLloyd:
Cool

AnthonyLloyd:
Still have the 7 failing tests. Changing to LF doesn't seem to fix.

haf:
Can you set up an appveyor.yml file to showcase the test errors?

AnthonyLloyd:
@haf

Sure will look to.

I also did a contention profile and think that the problem with the slow down when using many workers is down to assertTestFails and multiple full evals running. Still experimenting but this probably could be transformed into a normal test without using the full runner inside the test.

I may also pick up #68 later in the week if no one else has done it. I have an interest in what could be done later with FsCheck (scaling for stress testing, easy stdgen etc)

Issue #63, opened 2017-01-07
Change distribution error message


Krzysztof-Cieslak:
I don't think distribution error should display those part of message which are not not showing actual error for the case (so for example don't display `Missing elements from actual:` + empty line if there are no missing elements).

CC: @haf and @MNie

MNie:
@Krzysztof-Cieslak  I agree, also the same with continsAll? If there are not any elements it shouldn't display:
`missing/extra elements from/in 'actual'`
`{}`
?

PR #64, opened 2017-01-07
#63 change distribution error msg

MNie:
#63 

CC: @Krzysztof-Cieslak @haf

### DIFF
--- a/Expecto/Expect.fs
+++ b/Expecto/Expect.fs
@@ -293,9 +293,8 @@ let distribution (actual : _ seq)
 
   let missingElementsInfo, extraElementsInfo = 
     let incorrectElementsInfo elements is direction =
-      match List.isEmpty elements with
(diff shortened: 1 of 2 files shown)


Expecto/Expect.fs
```diff
@@ -290,17 +290,22 @@ let distribution (actual : _ seq)
     expected
     |> Map.toList
     |> List.map (fun element -> sprintf "%A: %d" (fst element) (snd element))
+
+  let missingElementsInfo, extraElementsInfo = 
+    let incorrectElementsInfo elements is direction =
+      match List.isEmpty elements with
```
haf: prefer `if` for booleans

MNie: @haf fix'd

haf:
Thanks!

PR #65, opened 2017-01-08
add Expect.floatClose with better accuracy test

AnthonyLloyd:
@haf,

Here is something I find useful for your consideration. It allows you to compare floats that are either large or small by combining an absolute and relative accuracy.

Its along the lines of https://docs.scipy.org/doc/numpy-1.10.0/reference/generated/numpy.isclose.html but better :).

You may want to look to deprecate `Expect.floatEqual`

### DIFF
--- a/Expecto/Expecto.fs
+++ b/Expecto/Expecto.fs
@@ -1114,19 +1114,20 @@ module Tests =
 
 
 type Accuracy =
-    {
-        absolute: float
-        relative: float
-    }
+  {
+    absolute: float
+    relative: float
+  }
 
 [<CompilationRepresentation (CompilationRepresentationFlags.ModuleSuffix)>]
 module Accuracy =
   let inline areCloseLhs a b = abs(a-b)
-  let inline areCloseRhs m a b = m.absolute + m.relative * (max (abs a) (abs b))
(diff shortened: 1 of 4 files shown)


haf:
This is a very nice contribution. I agree that generality is better than specificity.

AnthonyLloyd:
@haf rebased, added tests, docs and marked floatEqual as obsolete.

Expecto/Expecto.fs
```diff
@@ -1111,3 +1111,23 @@ module Tests =
...
+  }
+
+[<CompilationRepresentation (CompilationRepresentationFlags.ModuleSuffix)>]
+module Accuracy =
+  let inline areCloseLhs a b = abs(a-b)
+  let inline areCloseRhs m a b = m.absolute + m.relative * max (abs a) (abs b)
+  let inline areClose m a b = areCloseLhs a b <= areCloseRhs m a b
+
+[<AutoOpen>]
+module AccuracyAuto =
```
haf: What do you think of moving these module fields up into `Accuracy`?

AnthonyLloyd: Will do, makes sense.

PR #71, opened 2017-01-09
async docs, async pending/focused, closes #2

### LINKED ISSUE #2: Support async tests

AnthonyLloyd:
Also:
- minor change to fasterThan wording
- small change to parallelLimit (only need List.length workers if max is too high)

### DIFF
--- a/Expecto/Expecto.fs
+++ b/Expecto/Expecto.fs
@@ -88,6 +88,7 @@ module Async =
 
   /// Traverses the list of async, spawning them with Async.Start and
   let parallelLimit maxParallelism s =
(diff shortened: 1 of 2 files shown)


haf:
Thanks @AnthonyLloyd !

PR #73, opened 2017-01-10
Not FAKE stats for Expecto =)

sergey-tihon:


haf:
You see, expecto was very popular very quickly. ;)

PR #75, opened 2017-01-11
add a C# wrapper for convenience

baronfel:
This adds a new project for authoring a C# wrapper that handles nice interop around `IEnumerable<Test>` instead of `Test list` and `Task` instead of `Async<unit>` for our C# friends.  There is also a small sample test project to verify that things work.

Is this valued? If so, what needs to be done at this point to author packages?


haf:
I'm going to make the executive decision here and say that I want expecto to be F# only in its core repository, but I wouldn't mind if you publish this as a nuget.

If you use albacore like is done here you don't need to do anything to publish the new package. It's all automatic and you just run 'bundle exec rake release'.

forki:
Can't you write the same thing in F# and put it in the core repo? Then we would only have the test csproj - which I believe would be a good thing. I personally don't think writing tests in C# is useful in any way, but it's probably still a good thing to attract more people.

haf:
https://github.com/logary/logary/blob/feature/live-wpc-reconfigure/src/Logary/CSharp.fs

haf:
Perhaps you can take a leaf from the above link?

baronfel:
Yes that looks to be exactly what I need. I'll try to get to it today.

haf:
I'm closing this in the mean while. Tell me if you need any help.

Issue #72, opened 2017-01-10
Expecto logo


sergey-tihon:
Would love to hear opinions from fans
// @haf @forki @Krzysztof-Cieslak @MNie @AnthonyLloyd @cloudRoutine @adamchester 

![image](https://cloud.githubusercontent.com/assets/1197905/21821768/c965d092-d785-11e6-8564-5e8fa2bc38a5.png)
![image](https://cloud.githubusercontent.com/assets/1197905/21821640/647a24c6-d785-11e6-9f3c-1cb8ead57422.png)

Krzysztof-Cieslak:
:+100:

forki:
:kiss:

adamchester:
❤️

haf:
Hey @forki, surely this is not enterprise enough for you?

forki:
@haf the issue with the moot summary is that it looks weird. Even on modern console output. Nobody in the room saw the smiley. It just was weird.

haf:
;)

forki:
But that's OK. We iterate on things. We should also iterate on jokes

AnthonyLloyd:
👍 

I just see a series of boxes in powershell

haf:
@forki I'm worse myself; I've been an enterprise dev for 10+ years now and there's no end in sight. Enterprise is good money so I don't mind. Btw, we should definitely have a beer sometime, when are you coming to Stockholm next? :)

@sergey-tihon Concretely, I think the logo is too warm in its colours, with both brown and orange being reds. Also, do we want more Harry?

In any case, we don't have a logo just yet, so let's make this the one? Albacore picks out the path from the fsproj file on publish, so just add an `Icon` element with the url to the properties and it'll be used.

Krzysztof-Cieslak:
Could I join? ;]

haf:
Yes why don't you all in this thread come to Stockholm for a few days and let me show you around? And for great conversation.

AnthonyLloyd:
I'll be in Stockholm in the summer en famille for our hols.

sergey-tihon:
@haf what about scarf and cold colors?
![image](https://cloud.githubusercontent.com/assets/1197905/21844119/7fae5c48-d7fe-11e6-9265-960b64ad42e6.png)

haf:
@sergey-tihon When in doubt, use pastels. I do like the scarf =)

@AnthonyLloyd What dates?

AnthonyLloyd:
[...]

PR #78, opened 2017-01-12
Logo

sergey-tihon:
Fix for #72

@haf what also should I do?

>Albacore picks out the path from the fsproj file on publish, so just add an Icon element with the url to the properties and it'll be used.

Does it mean that image will be compiled as resource into Expecto dll?


haf:
No, it's just an XML-element.

Could you please add the vector version?

sergey-tihon:
I can do something like
```xml
 <PropertyGroup>
    <ApplicationIcon>https://github.com/haf/expecto/raw/master/docs/expecto-logo-small.png</ApplicationIcon>
  </PropertyGroup>
```
but i am not sure that it is correct

haf:
Yes, but name it like the [nuspec](https://docs.microsoft.com/en-us/nuget/schema/nuspec) – `<IconUrl>...</IconUrl>`

Do you have the file as an `.ai`, otherwise I could vectorise it after merging. Thank you for creating a logo for the project!

haf:
Docs [here](https://github.com/Albacore/albacore/wiki/nugets_pack)

sergey-tihon:
@haf done, please update proj file by yourself

haf:
Super, thank you!

PR #82, opened 2017-01-16
FsCheck messages, focus and pending methods, closes #68

### LINKED ISSUE #68: FsCheck prints to stdout – make it print to `Logger`
Our [FsCheck.fs](https://github.com/haf/expecto/blob/master/Expecto.FsCheck/FsCheck.fs) uses the `Config.Default.Runner` on [line 39](https://github.com/haf/expecto/blob/master/Expecto.FsCheck/FsCheck.fs#L39-L41).

Change the FsCheck integration to use a few tricks from the [XUnit integration](https://github.com/fscheck/FsCheck/blob/master/src/FsCheck.Xunit/PropertyAttribute.fs) in the FsCheck repo. Amongst things ensure we have our own [Expecto runner](https://github.com/fscheck/FsCheck/blob/master/src/FsCheck.Xunit/PropertyAttribute.fs#L19-L29) that properly logs and collates input so that input to the test case can be printed/logged on failure.

/cc @mausch @kurtschelfthout

AnthonyLloyd:
Focused test property takes StdGen. Will do docs if ok.

### DIFF
--- a/Expecto.FsCheck/FsCheck.fs
+++ b/Expecto.FsCheck/FsCheck.fs
@@ -1,55 +1,85 @@
 ﻿namespace Expecto
 
 open System
-open global.FsCheck
+open FsCheck
+open Expecto
 
 [<AutoOpen; CompilationRepresentationAttribute(CompilationRepresentationFlags.ModuleSuffix)>]
 module ExpectoFsCheck =
-  open Expecto
-  open Expecto.Helpers
-  open Expecto.Impl
-  open global.FsCheck
-  open global.FsCheck.Runner
 
-  let internal (|Ignored|_|) (e: exn) =
-    match e with
-    | :? IgnoreException as e -> Some e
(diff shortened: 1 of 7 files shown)


haf:
I think it looks good from a quick glance.

haf:
So does `ftestProperty` only run the focused test? In my understanding it's in this new API just a way to specify the seed explicitly; shouldn't we name it something else then?

AnthonyLloyd:
Yes its focused and seeded.

Do you think we need others versions? I thought if you need to focus you probably need to seed also? Easy to put in others if not.

haf:
Fair enough, thank you! :)

haf:
I'll try them out myself and put some more comments here if needed.

AnthonyLloyd:
Cool. Thanks

haf:
Looks really good, thanks.

PR #85, opened 2017-01-17
fixes #84 parallel bug

### LINKED ISSUE #84: Update to Expecto 3.0.2 broke our build
Something rather important broke between 3.0.1 and 3.0.2 and that breaks our build.
Is there anything known to be breaking? BTW: 3.1 is broken for us as well.

I diffed and didn't see anything obvious - apart from that huge change in the lock file.

/cc @theimowski @haf @Krzysztof-Cieslak

AnthonyLloyd:
Also added two tests for parallel = false.

Could not reproduce the hanging though.


haf:
Thanks!

PR #88, opened 2017-01-24
Default message texts for isTrue and isFalse

isaacabraham:
Add friendly messages for isTrue and isFalse. Also reword notEqual.


haf:
Thank you @isaacabraham !

Issue #92, opened 2017-02-07
Float assertions


haf:
PR this

```fsharp
module Expect =
  /// Expect the passed float to be a number.
  let isNotNaN f format =
    if Double.IsNaN f then Tests.failtestf "%s. Float was the NaN (not a number) value." format

  /// Expect the passed float not to be positive infinity.
  let isNotPositiveInfinity actual format =
    if Double.IsPositiveInfinity actual then Tests.failtestf "%s. Float was positive infinity." format

  /// Expect the passed float not to be negative infinity.
  let isNotNegativeInfinity actual format =
    if Double.IsNegativeInfinity actual then Tests.failtestf "%s. Float was negative infinity." format

  /// Expect the passed float not to be infinity.
  let isNotInfinity actual format =
    isNotNegativeInfinity actual format
    isNotPositiveInfinity actual format
    // passed via excluded middle

  /// Expect the passed string not to be empty.
  let isNotEmpty (actual : string) format =
    Expect.isNotNull actual format
    if actual.Length = 0 then Tests.failtestf "%s. Should not be empty." format
```

MNie:
`if actual.Length = 0 then Tests.failtestf "%s. Should not be empty." format`

Why not:
```
let isNotEmpty (actual: string) format =
    if String.IsNullOrEmpty actual then Tests.failtestf "%s. Should not be empty." format
```

Also what about isNotWhitespace?

```
let isNotWhitespace (actual: string) format =
  if String.IsNullOrWhiteSpace actual then Tests.failtestf "%s. Should not be whitespace." format
```

CC: @haf

haf:
Mostly because with the code as I wrote it, you'll get a particular error if it's null and another one if it's empty, allowing you to take note which case it is.

Is not white space should be constructed in a similar fashion, initially asserting it's non-empty and then asserting it's not white space.

MNie:
@haf makes sense for me right now 👍

PR #93, opened 2017-02-07
Issue 92 float assertions

MNie:
As mentioned in issue: #92 

CC: @haf


MNie:
@haf  `isnotwhitespace` added.

haf:
Thank you @MNie !

PR #95, opened 2017-02-11
#92 update readme

MNie:
Update readme to include new functions.
#92 #93 

CC: @haf


haf:
Thank you!

Planning on releasing the next major later today.

PR #96, opened 2017-02-11
increase sleep in deadlock

AnthonyLloyd:
Can't reproduce on my machine with a stress test. Build machine must be quite different.
Its down to trying to create a deadlock in two tests. Not so easy to guarantee.
I've put up the sleep time in the deadlock from 10ms to 100ms. I think this should do the trick.
Its not an issue with the functionality just the ability to create a failing test.
Let me know if you see it again.

### DIFF
--- a/Expecto.Tests/Tests.fs
+++ b/Expecto.Tests/Tests.fs
@@ -752,6 +752,7 @@ let close =
 
   ]
 
+open System.Threading
 
 [<Tests>]
 let stress =
@@ -771,8 +772,8 @@ let stress =
       ]
 
     let deadlockTest() =
-      let waitOne = new System.Threading.ManualResetEventSlim()
-      let waitTwo = new System.Threading.ManualResetEventSlim()
+      let waitOne = new ManualResetEventSlim()
+      let waitTwo = new ManualResetEventSlim()
       let lockOne = new obj()
       let lockTwo = new obj()
       testList "deadlock" [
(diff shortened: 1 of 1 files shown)


AnthonyLloyd:
Also now using a readerwriterlock to improve this.

Expecto.Tests/Tests.fs
```diff
@@ -771,24 +771,22 @@ let stress =
...
       testList "deadlock" [
         testCaseAsync "case A" <| async {
-          lock lockOne (fun () ->
-            Thread.Sleep 10
-            lock lockTwo (fun () ->
-              ()
-            )
-          )
+          lockOne.EnterReadLock()
+          Thread.Sleep 100
```
haf: May I suggest that you fire off a semaphore in A, that you WaitOne on in B? That way you can easily sequence it so that both A and B have taken the read lock before they try to acquire the write lock.

AnthonyLloyd: Thanks. Trying with reset events.

AnthonyLloyd:
Sorry it seems to be erroring now (returning 10). I'm just going out so will have to look at later. You can release with these tests pending if you like and I will fix properly.

AnthonyLloyd:
@haf looks like the simple answer is to make sure there are a number of workers (8) running. All three versions in travis passed and I've run a 10 min stress test on my machine.

haf:
Nice! How useful that someone had PRed a stress test feature! ;)

PR #97, opened 2017-02-11
remove reader writer locks

AnthonyLloyd:
Still failing! I've removed the reader writer locking and reduced the sleep. Hoping its only the number of workers and a small sleep needed.

### DIFF
--- a/Expecto.Tests/Tests.fs
+++ b/Expecto.Tests/Tests.fs
@@ -792,7 +792,7 @@ let stress =
           ) ()
         }
       ]
-//touch to test in travis
+
     let sequencedGroup() =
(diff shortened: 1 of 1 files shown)


haf:
Soon you've tried everything! ;)

AnthonyLloyd:
:) I've got it working now. It does need to be sequenced though as other tests are competing. Amazing its so hard to force a deadlock consistently.

Will do one more them squeeze them down.

AnthonyLloyd:
@haf All done. Passed 12 travis build jobs and 3 more if these ok.

haf:
Thank you @AnthonyLloyd!

PR #98, opened 2017-02-16
minor --version bug fix and paket.local support

AnthonyLloyd:
Spotted a bug with the printing of version getting mixed up while using Expecto.

Also add FAKE Target to make paket.local easy as follows:

nuget Expecto -> git file:///C:\Users\Ant\src\expecto master build:"build.cmd CreateNuGets", Packages: /build/pkg/


haf:
Thanks!

PR #115, opened 2017-02-23
C# Interop module

baronfel:
Reopening #75 to hopefully get it over the finish line this time.

I've made a module inside the Expecto code that provides the necessary members, and there's now a C# test project in the solution that is using the new members.  I'm mostly concerned with the use case from the C# side.  I'd rather not have to `using static` the `Expecto.CSharp.Runner` module, so does anyone have an idea of how we could expose this in a more palatable way?  If the consumer can be convinced to using static then it flows pretty nicely.


Expecto/Expecto.fs
```diff
@@ -1565,4 +1564,36 @@ module Accuracy =
...
+
+[<AutoOpen; Extension>]
+module Runner =
+  open Expecto
+  open Expecto.Impl
+  open System.Collections.Generic
+  open System.Threading.Tasks
+
+  type Async with
+    static member AwaitTask(t : Task) = Async.AwaitIAsyncResult(t, -1) |> Async.Ignore
```
haf: This implementation is wrong

Expecto.Tests.CSharp/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Reflection;
```
haf: Auto generated?

Expecto/Expecto.fs
baronfel: There is a native implementation in F# 4.0 if we can upgrade the FSharp.Core dependency from 3.1.2.5.  Would you be open to this?

Expecto.Tests.CSharp/Properties/AssemblyInfo.cs
baronfel: Ah, yes. Will remove asap.

baronfel:
Ok, removed the autogen AssemblyInfo from the C# project.  If you're ok with upgrading the F# Core reference to 4.0.0.1 for the native untyped-Task-Await function, I've got that commit ready and successfully building/tested locally as well.

haf:
Go ahead.

baronfel:
Some feedback from Darrell Miller on twitter:

<blockquote class="twitter-tweet" data-lang="en"><p lang="en" dir="ltr"><a href="https://twitter.com/ChetHusk">@ChetHusk</a> The only unusual thing is for a static factory method to not have &lt;verb&gt;&lt;noun&gt; format. /cc  <a href="https://twitter.com/sforkmann">@sforkmann</a> <a href="https://twitter.com/ReidNEvans">@ReidNEvans</a></p>&mdash; Darrel Miller (@darrel_miller) <a href="https://twitter.com/darrel_miller/status/834799549780738052">February 23, 2017</a></blockquote>

Also, @forki pointed me to the prior [Fuchu C# samples](https://github.com/mausch/Fuchu/blob/master/Fuchu.CSharpTests/Program.cs), perhaps worth porting those over/mimicing that API?

haf:
@baronfel You're the author; I'm not going to try to sell Expecto to C# people personally.

I also have a suggestion; at the end of this pull request, I'm going to make you a committer so that you can change the C# side of things rapidly to suit your C# project requirements. How does that sound?

haf:
This change will go into version 5 since you've bumped the F# Core dependency. You could get it out in a minor version if you revert that change and copy the implementation into Expecto.

haf:
Gentle ping as to query your progress?

baronfel:
Thanks for the poke, I've reverted the FSharp.Core upgrade and implemented a new AwaitTask implementation that uses a TaskCompletionSource to signal completion, so we can use the typed-Task AwaitTask implementation in FSharp.Core.  I couldn't lift the implementation directly from FSharp.Core because it relies on some rather hairy internal functions that I also didn't want to life entirely.  Does this implementation mesh with what you're expecting?  I'm also going to rebase this to account for merge conflicts.

As to the surface area of the API, I'm more than happy with what's exposed. I think there's enough there to instruct anyone who might want to add more members how to do it easily.

haf:
Yes, that's much better.

baronfel:
Things looks mostly shipshape. I turned on Redirects as sort of a general Paket best practice which is where some of the churn comes from.  `Expecto.fsproj` looks rather nasty presumably because of a whitespace difference that we could take care of, and it looks like my version of msbuild (the one shipped with mono 4.8) wants to rewrite some of the casing/GUID formats in the project files, though that could also be Xamarin Studio or VS for Mac doing that.  Let me know if any of that is a blocker.

haf:
Thank you @baronfel – you are now proud maintainer of Expecto.CSharp! https://github.com/haf/expecto/blob/master/Expecto/Expecto.fs#L1598 :)

baronfel:
A dubious honor, to be sure :D

PR #116, opened 2017-02-23
Add async computation expression plus tests

AnthonyLloyd:


README.md
```diff
@@ -208,8 +208,9 @@ they are using `Expecto.BenchmarkDotNet` for performance tests.
 
 ### Normal tests
 
- - `testCase : string -> (unit -> unit) -> Test`
  - `test : string -> TestCaseBuilder`
+ - `testAsync : string -> TestAsyncBuilder`
```
haf: Could you add a summary doc at the end of each line in this list?

Expecto/Expecto.fs
```diff
@@ -1346,6 +1346,31 @@ module Tests =
...
+    member __.Combine(p1, p2) = async.Combine(p1, p2)
+    member __.TryFinally(p, cf) = async.TryFinally(p, cf)
+    member __.TryWith(p, cf) = async.TryWith(p, cf)
+    member __.Run f =
+      match focusState with
+      | Normal -> testCaseAsync name f
+      | Focused -> ftestCaseAsync name f
+      | Pending -> ptestCaseAsync name f
+
+  let inline testAsync name =
```
haf: Docs here

Expecto/Expecto.fs
```diff
@@ -1346,6 +1346,31 @@ module Tests =
...
+    TestAsyncBuilder (name, Normal)
+  let inline ftestAsync name =
```
haf: Here

Expecto/Expecto.fs
```diff
@@ -1346,6 +1346,31 @@ module Tests =
...
+    TestAsyncBuilder (name, Focused)
+  let inline ptestAsync name =
```
haf: And here

Expecto/Expecto.fs
```diff
@@ -1346,6 +1346,31 @@ module Tests =
   let inline ptest name =
     TestCaseBuilder (name, Pending)
 
+  type TestAsyncBuilder(name, focusState) =
```
haf: Docs here

haf:
Forgot to ask you: how does this PR interfere with the Cecil workings?

AnthonyLloyd:
That's a good point. I'm not sure if the test CE come out in the current cecil output. I know ones that resolve to `Async test` don't as I couldn't write the code to do it and there a placeholder "Unknown Async".

@Krzysztof-Cieslak Do you think there is a way to locate async's?

haf:
It may also be apt to ask @jbevain to his input to whether Cecil could support these changes?

Expecto/Expecto.fs
```diff
@@ -1,1351 +1,1352 @@
-namespace Expecto
```
haf: You just added Windows-newlines everywhere. I recommend installing the EditorConfig plugin.

AnthonyLloyd: Ah odd. Just did it in the web editor. Will redo from code.

Krzysztof-Cieslak:
> Do you think there is a way to locate async's?

To be fair I have no idea  :-(

AnthonyLloyd:
@haf on your how many testCase q. I tried to leave in a mix but put the test CE first. The second is a testCase and they are generally mixed further down.

haf:
I'm leaving it up to you to merge at will pending how much you value/test the feature of finding the line numbers from Ionide and pending the answer from jbevain.

AnthonyLloyd:
Its worth looking into. If we promote test CE and it doesn't work for Cecil then we'd undo the Ionide work.

I'm willing to take a look but I don't get any output from Cecil when I run the code. Does it need some setup I'm missing? Interesting for me to learn about it anyway.

jbevain:
@haf what are you looking to achieve with Cecil?

Krzysztof-Cieslak:
@jbevain. We're using Cecil to find source file location of executed tests on runtime to allow better editor tooling integration - https://github.com/haf/expecto/blob/master/Expecto/Expecto.fs#L1140-L1210

Krzysztof-Cieslak:
As far as I understand, current method works well if tests are defined normally but for some reason fails for async tests? @AnthonyLloyd knows more

Krzysztof-Cieslak:
Also, looking at this code again - the problem may lay before Cecil is called, in part using normal .Net reflection.

jbevain:
@Krzysztof-Cieslak cool! I'm referring to:

> It may also be apt to ask @jbevain to his input to whether Cecil could support these changes?

I don't have the context. I'd be glad to help you using Cecil and answering any question you have, but you'll need to explain what you want to do, what doesn't work, and what changes you mean :)

AnthonyLloyd:
Its in this section https://github.com/haf/expecto/blob/master/Expecto/Expecto.fs#L1140-L1174

The code locates a function but I'm not sure what to do to locate the code for Async<unit> or the AsyncFsCheck.

```
type TestCode =
  | Sync of stest: (unit -> unit)
  | Async of atest: Async<unit>
  | AsyncFsCheck of testConfig: FsCheckConfig option *
                    stressConfig: FsCheckConfig option *
                    test: (FsCheckConfig -> Async<unit>)
```

haf:
LGTM: no breaking changes and this discussion can still be held.

haf:
Thank you Anthony!

Issue #107, opened 2017-02-21
Order of parameters in Expecto module


smoothdeveloper:
I was looking at samples https://github.com/haf/expecto/blob/master/Expecto.Sample/Expecto.Sample.fs and was wondering if there was value of making the API friendlier to piping or if the choice of order was made to discourage piping.

A simple example

```fsharp
Expect.isTrue subject "I compute, therefore I am."
```

It would make sense in certain cases to be able to write it that way instead:

```fsharp
subject |> Expect.isTrue "I compute, therefore I am."
```

That is passing the actual as last parameter.

That would be for sure a breaking change to the API, but maybe worth consideration if people using the library are looking forward more piping to the assertion functions.

haf:
@smoothdeveloper How about a PR with the alternative ordering in a separate module? That way people can choose what they prefer.

PR #119, opened 2017-02-24
Add a Expecto.Flip.Expect module

smoothdeveloper:
Proposal for #107

The main thing I'm concerned about are the isGreaterThan & co, and maybe the contains; others seem to read well.

Note that I renamed `format` into `message` as that parameter is not a format (from my understanding) but the failure message. I can rename those in the plain module or make them `format` here.

### DIFF
--- a/Expecto.Sample/Expecto.Flip.Sample.fs
+++ b/Expecto.Sample/Expecto.Flip.Sample.fs
@@ -1,0 +1,92 @@
+﻿module Expecto.Flip.Sample
+
+module Utils =
+  let repeatN n f a =
+    let mutable v = f a
+    for n in 2 .. n do
+      v <- f a
+    v
+  let repeat10 f a = repeatN 10 f a
+  let repeat100 f a = repeatN 100 f a
+  let repeat1000 f a = repeatN 1000 f a
+  let repeat10000 f a = repeatN 10000 f a
+
+open Utils
+open Expecto
+open Expecto.Flip
+
+[<Tests>]
+let tests =
+  testList "flipped samples" [
+    testCase "universe exists (╭ರᴥ•́)" <| fun _ ->
+      let subject = true
+      subject |> Expect.isTrue "I compute, therefore I am."
+
+    testCase "when true is not (should fail)" <| fun _ ->
+      let subject = false
+      subject |> Expect.isTrue "I should fail because the subject is false"
+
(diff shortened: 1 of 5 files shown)


Expecto.Sample/Expecto.Flip.Sample.fs
```diff
@@ -0,0 +1,92 @@
...
+    test "Should not be close enough (should fail)" {
+      let actual, expected = 41.562, 41.563
+      actual |> Expect.floatClose "Should be close within 5 sig figs (approx)" Accuracy.medium expected
+    }
+
+    testCase "1 is > to 0" <| fun () ->
+      1 |> Expect.isGreaterThan "1 > 0" 0
+    
+    testCase "0 is > to 1 (should fail)" <| fun () ->
+      0 |> Expect.isGreaterThan "1 > 0" 1
```
haf: Doesn't this look really funny for this assertion? "0 ... is greater than ... 1"

Expecto/Flip.Expect.fs
```diff
@@ -0,0 +1,138 @@
+﻿module Expecto.Flip.Expect
```
haf: Remove CRLF please

haf:
Looks good. Would you like to rename `format` to `message` in the main `Expect` module as well?

Expecto.Sample/Expecto.Flip.Sample.fs
smoothdeveloper: I'd say worrisome even (when not piped), wanted to make it all just so we can decide what it should be. 

I'm considering a tuple `(a, b)` would be the correct version in Flip module, What do you think?

haf: Could you expand on that with an example please?

smoothdeveloper:
```fsharp
(1,0) |> Expect.isGreaterThan "1 > 0"
```

haf: Sure, for the comparison ops.

haf:
Looks great. Remember to ping me when you push and are ready for me to review, as ordinary pushes don't cause notifications.

Ready to merge?

smoothdeveloper:
@haf sorry will remember I didn't know.

Yes I think it is ready to get some use in the wild if people like it!

Issue #94, opened 2017-02-11
Allow specifying the Assembly to run tests in


mtnrbq:
I've added a new function to my fork of Expecto - runTestsInDifferentAssembly 
My particular use case is where I have the tests in a F# class library, but I need to run the tests from within a c# exe.

It's meeting my need, happy to contribute a PR if you believe it'd be of general use.

```FSharp
  /// Runs tests in a passed in assembly with the supplied command-line options.
  /// Returns 0 if all tests passed, otherwise 1
  let runTestsInDifferentAssembly config args (assembly:Assembly) =
    let config = { config with locate = getLocation (assembly) }
    testFromDifferentAssembly assembly
    |> Option.orDefault (TestList ([], Normal))
    |> runTestsWithArgs config args
```

haf:
I think this for next version:

Extract an `Assembly` param from https://github.com/haf/expecto/blob/master/Expecto/Expecto.fs#L1541-L1542

Rename https://github.com/haf/expecto/blob/master/Expecto/Expecto.fs#L1540 to `runTestsInAssembly`

Then there's your function. Then below:

```fsharp
let runTestsInThisAssembly config args =
  runTestsInAssembly config args (Assembly.GetEntryAssembly())
```

Now we have both features and only one line more of code.

haf:
Hi there. I'll be closing this issue tomorrow due to the lack of a PR/response.

MecuSorin:
Hey Henrik, I think we can shoot 2 bunnies with a single bullet: use this to exercise the tests from Expecto.Focused.Tests (we can make Expecto.Tests to reference Expecto.Focused.Tests, make the focused test from Expecto.Focused.Tests to fail, and the normal test to succeed) run the tests from Expecto.Focused.Tests and check that fails. No need then to change the build.fsx and rakefile.
As a contra argument: this scenario is giving too much complexity for my taste.  What do you think?

Can I make a PR on this particular issue? 
I don't have an idea of proper testing the code now.

haf:
I think it would make sense to test this by referencing your project. Go ahead and PR it. The return value of the main test suite should just be the sum of the two calls to test tests.

mtnrbq:
Hi, I've made the PR for the change in Expecto.fs - I have been hung up on how best to add testing code. This is my first PR, any advice or guidance for future appreciated

haf:
@mtnrbq See comment https://github.com/haf/expecto/issues/94#issuecomment-279134364

[...]

PR #122, opened 2017-03-02
Allow specifying the Assembly to run tests in #94

mtnrbq:


mtnrbq:
Redoing to include fixes for test suite

Issue #127, opened 2017-03-11
Sequence assertions


haf:
Add `Expect.isEmpty` and `Expect.isNonEmpty`.

Also add `val hasCountOf: subject:seq<'a> -> expected:uint32 -> selector:('a -> bool) -> message:string` that counts the occurrences of the found value in the sequence and ensures the number of found such equal the expected.

PR #128, opened 2017-03-11
Feature/127/isempty

MNie:
#127 
CC: @haf

### DIFF
--- a/Expecto/Expect.fs
+++ b/Expecto/Expect.fs
@@ -178,10 +178,11 @@ let isEmpty (actual) message =
 let isNonEmpty (actual) message =
   if Seq.isEmpty actual then Tests.failtestf "%s. Should be empty." message else ()
 
-/// Expect the passed sequence to contains expected number of elements.
-let hasCountOf (actual : _ seq) (expectedLength : int32) message =
-  if Seq.length(actual) = expectedLength then ()
-  else Tests.failtestf "%s. Should be of length: %s, but was: %s" message expectedLength actual.Length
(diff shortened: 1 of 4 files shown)


Expecto/Expect.fs
```diff
@@ -170,6 +170,19 @@ let isNotWhitespace (actual : string) message =
...
+/// Expect the passed sequence to be empty.
+let isEmpty (actual) message =
+  if not (Seq.isEmpty actual) then Tests.failtestf "%s. Should be empty." message
+
+/// Expect the passed sequence to be not empty.
+let isNonEmpty (actual) message =
+  if Seq.isEmpty actual then Tests.failtestf "%s. Should be empty." message
+
+/// Expect that the counts of the found value occurrences in sequence equal the expected.
+let hasCountOf (actual : _ seq) (expected : uint32) (selector : _ -> bool) message =
```
haf:
```
    let len =
      xs |> Seq.fold (fun acc x -> if selector x then acc + 1u else acc) 0u
    Expect. ...
```

Would probably be more efficient.

haf:
Nice!

Expecto/Expect.fs
MNie: Fix'd :)
CC: @haf

haf:
Sorry, I forgot; do you think you could add these to the README as well?

MNie:
@haf but I did that [in this commit](https://github.com/haf/expecto/pull/128/commits/860f3d06447a3b39e3f213ce87c8916adff2a4b6) :)

haf:
True, don't mind me ;)

PR #129, opened 2017-03-13
netcore: run `dotnet test` on windows builds

adamchester:
And update Paket and Mono.Cecil


adamchester:
This is ready to go into `netcore` branch if we're OK with this.

haf:
Sure, we're OK. Looks good!

PR #130, opened 2017-03-19
fixed typo in first example in Readme

stelim:
Small typo, which leads to an exception, because "Hello **w**orld" is not equal to "Hello **W**orld".
-->  Expecto.AssertException: The string should equal.


haf:
Thanks

PR #131, opened 2017-03-22
stress deadlock test only for processorcount > 2

AnthonyLloyd:


haf:
Great! Stable tests make for happy people.

Issue #132, opened 2017-03-27
Add `Expect.isMatch actualString regexString`


haf:
And another `Expect.isRegexMatch : string -> Regex -> unit`

Aim: let the user expect a string to match a regex.

Also add to `Expect.Flip`.

Maybe this?

```fsharp
module Expect =
  let isMatch actual rstring message =
    match Regex.``match`` rstring actual with
    | None ->
      Tests.failtestf "%s. Expected %s to match /%s/" message actual rstring
    | Some x ->
      x
```

Plus the rest + docs?

MNie:
@haf isNotRegexMatch also?

haf:
Yes, that makes sense. And perhaps one that takes a callback that takes the matched groups.

MNie:
@haf 
>And perhaps one that takes a callback that takes the matched groups.`

You mean something like this?

    let isRegexMatchGroups actual pattern doSomethingWithMatch message =
        let groups = Regex.Match(actual, pattern).Groups
        match doSomethingWithMatch groups with
        | Some x -> ()
        | None -> Tests.failtestf "%s. Expected %s to match pattern: /%s/" message actual pattern

haf:
Thinking about it, perhaps the best version is to return the groups if there's a match. Let's merge what you've already done, it's good enough for now, and we can ponder the groups ;)

haf:
Fixed with #133

PR #133, opened 2017-03-27
#132 ismatch and is not match

MNie:
#132 
- isRegexMatch
- isNotRegexMatch


haf:
The regex infix should only be there when you pass a System.Text.RegularExpressions.Regex. Otherwise, leave out regex and pass just the string containing a regex expression.

haf:
Looks great so far.

haf:
Thank you @MNie – excellent.

PR #134, opened 2017-03-28
#132 [WIP] isGroupsMatch and isRegexGroupsMatch

MNie:
#132 
Continuation of #133  + some fixes to code style (spaces before `:` and rename of parameters to match what they are in fact (`pattern -> regex`)) 
@haf what do you think? Is it what you `expecto`? :)


haf:
👍 Yes, this is nice. I think we can tighten up the functions a bit by providing an API that selects groups and (while dotting/indexing the way there) also asserts the group actually exists – the PR as it stands will throw ItemNotFoundException or similar on the indexing operation if it fails. In that case, we might as well just return the groups.

I waited a bit for this PR because I needed to think. Have you seen the regex type provider? Wouldn't it be cool is we could provide the parsed regex as a callback value to be asserted on?

MNie:
@haf I checked RegexTypeProvider it looks cool!
So suppose we have a type like this:
`type RegexType = Regex< @"(?<ExactMatch>^\d{3})" >`
 You prefer to create a function (isMatchWithCallback or something) which will accept function like: `RegexType -> bool` or `string -> bool` where string in fact will be a: `RegexType.ExactMatch`?

haf:
Hi @MNie – I do get notified about your comments here. ;) Can we leave this feature for a while, please? I want to try the new regex support myself, first; before deciding more regex PRs. I'm not closing it, nor am I saying this is not what we want, but I need to explore the domain a bit.

MNie:
@haf no problem :).

AnthonyLloyd:
@MNie I'm think what to do with this PR. I'm minded to merge. Are there any suggested changes that make sense given than once its in its harder to change.

MNie:
@AnthonyLloyd I think rather not, only merge with current master I will do that in minutes.

MNie:
@AnthonyLloyd merged with master :)

AnthonyLloyd:
@MNie Cool

AnthonyLloyd:
@MNie Thanks!

Issue #139, opened 2017-04-13
.net core template


MNie:
When migration to .net core is done #1 . 
It would be great to have an ability to create an Expecto .net core project by typing:
`dotnet new -i Expecto`
Like it is done for [Fable, Freya and so on.](https://github.com/dotnet/templating/wiki/Available-templates-for-dotnet-new)

CC: @haf

MNie:
Closing since template was created and docs were updated.

haf:
:1st_place_medal:

PR #143, opened 2017-04-19
.net core template information

MNie:
#139 
Add information to README about .net core template.
I don't know where to place it in README, so I add `.net core support` section after `Installing` section(?)
What more should be in this section?

CC: @haf


haf:
The picture you created, a sample command line and a link to your blog entry perhaps?

cloudRoutine:
Shouldn't the templates be added as part of the nuget package so that the dotnet cli can automatically detect and use them? This way when you add expecto to a project you'll be able to use the templates in that project automatically.

Isn't that how the new system is supposed to work?

MNie:
Haf: ok I will edit pr tommorow.
Cloudroutine: It is a nuget package in fact. You mean that expecto.template should be in the same repository as Expecto? I create it in separate repository because template for for example freya is also in different repository. Or You mean something different?

cloudRoutine:
I mean the template would be included in the `content` folder of expecto's `.nupkg` it might need an adidional nuspec next to it there or some other config file to indicate that it's a template, but the dotnet cli should recognize it then. 

I'm a little fuzzy on all the specifics I read about it somewhere but never bothered trying to set it up

MNie:
@cloudRoutine nuspec indicate that this is a template. The question is, wouldn't it provide some `mess` in Expecto package and repository?

cloudRoutine:
maybe i'll look into what it'll take to bundle the templates in the nupkg with the lib for cli consumption some other time 
@MNie you don't have to worry about it

haf:
Thank you @MNie !

PR #155, opened 2017-05-04
Add isOk and isError expectations

### LINKED ISSUE #62: Add isOk and isError expectations
Similar to `isChoice1Of2` and `isChoice2Of2`, it would be nice to have expectations that test for the two "branches" of the [`Result` type](https://github.com/fsharp/fslang-design/issues/49).

rmunn:
These test the Result type from F# 4.1, so they will only be available in Expecto versions 5.x and above.

This will fix #62.


haf:
@rmunn Thank you!

Issue #156, opened 2017-05-11
Expect.all and Expect.exists


haf:
```
allEqual: actual:#seq<'thing> -> thing:'thing -> message:string
all: actual:#seq<'thing> -> asserter:('thing -> bool) -> message:string
exists: actual:#seq<'thing> -> asserter:('thing -> bool) -> message:string
```

 - [x] allEqual
 - [x] all
 - [x] exists
 - [x] Expect.Flip
 - [x] Update README with docs on the above
 - [x] XMLDOC above each function
 - [x] Specify in the failure message, what iteration count you failed at, what the last seen item was and call/print a stubbed delta/diff function on the `'thing`, printing what differs between the expected and the found (that returns `false` from the `asserter`)

With "stubbed diff function", I mean a structural comparison, similar to what #6 would check for, but for a single object/type.

Example usage with Expect.Flip opened:

<img width="503" alt="screen shot 2017-05-11 at 10 39 54" src="https://cloud.githubusercontent.com/assets/193115/25940365/379dded8-3636-11e7-8d21-590ab4dd0d92.png">

haf:
Ping @gentauro and @MNie – if you're interested ;)

MNie:
@haf I could do that. I will provide pr with that tommorow if it's not a problem :)

haf:
Done! Thanks to MNie

PR #157, opened 2017-05-12
#156 add exists function

MNie:
#156 
`exists` function. 
`all` function, without `delta/diff function`
Error msg for `all` looks like this:
image
Is it ok? When `delta/diff function` would be implemented it could be changed to `Element at index %d which is equal to %A, has following difference %A"?
CC: @haf

### DIFF
--- a/Expecto.Tests/Tests.fs
+++ b/Expecto.Tests/Tests.fs
@@ -659,6 +659,19 @@ let expecto =
         ) |> assertTestFails
       ]
 
+      testList "#all" [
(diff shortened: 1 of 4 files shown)


haf:
"Some elements don't satisfy ..."

haf:
Yes, that sounds good; when the diff is there; then yes we change the message.

haf:
Thank you @MNie ! I appreciate your PRs and hope they continue coming.

PR #159, opened 2017-05-13
Use Seq/List item instead of Seq/List nth

MNie:
Remove all usages of List/Seq nth and replace them by List/Seq item since nth is deprecated.
So we don't see these warnings:
image

`Note: In F# 3.x, nth is defined with inconsistent signatures for Array and List. The proposal above replaces nth by item and would eventually deprecate nth (with a message to say 'please use Seq.item'. It also adds a corresponding tryItem. Both item and tryItem would take the integer index as the first parameter.` [source](https://visualfsharp.codeplex.com/wikipage?title=Status)

CC: @haf


haf:
Thank you!

PR #167, opened 2017-05-22
Updated with sample of custom user data

kunjee17:
Sample added to test thing using custom generated sample data.


haf:
Thank you @kunjee17 !

PR #166, opened 2017-05-18
use dotnet pack and nuspec instead of nuget client

enricosada:
To workaround nuget.commandline 3.5 who doesnt support prerelease deps as warning (instead of error)

fix https://github.com/haf/expecto/issues/165

In practice:

i didnt find a prerelease of nuget.commandline 4.0, so i used dotnet sdk, because anyway is required for final package.

`dotnet pack` = `dotnet msbuild /t:Pack` accept a nuspec file as property, instead of generate it from the msbuild project file itself (the common scenario for fsproj/csproj)

I added a project file `tools/nuget-workaround/nugetclient.proj` to be able to just run `Pack` target. like for `dotnet new` templates.

as a note, nuspec is just a temporary file now with new sdk. all props (minus owner) can be set as normal msbuild props and package generated with `Pack` target

TODO

- [x] win
- [x] osx/unix add bash script, alike of `tools/nuget-workaround/nugetclient.bat` to just invoke `dotnet msbuld` with passed arguments

PR #169, opened 2017-05-23
fix travis

enricosada:


enricosada:
@haf i need to wait travis, but probably is this

enricosada:
cherry-picked in https://github.com/haf/expecto/pull/166

PR #175, opened 2017-05-30
[WIP] Overhauled the project file structure and build system, and ported all three libraries to .NET Standard

### LINKED ISSUE #174: Add .NET Standard support for Expecto.FsCheck
Both Expecto and FsCheck now support .NET Standard. I think it's time to 🌉 the 🕳 between these two excellent projects.

teo-tsirpanis:
__NOTE: While the Pull request looks stable, it is under progress, as more features are added.__

This (kinda big) pull request addresses many things. Specifically:

- [x] Converts all project files to the new fsproj. This allows targeting both .NET Framework and .NET Standard from a single file. The project files are much more lightweight, easy-to-navigate and not changed every time a `paket update` is run.
- [x] Ported Expecto.FsCheck and Expecto.BenchmarkDotNet to .NET Standard (also closes #174).
- [x] The build system has been completely rewritten from Ruby's Rake to [FAKE](https://fake.build). This allows the developers to work completely in F# without installing unneccesary depedencies (I could not figure out how to install the ruby gems to my machine), and makes the project open to developers that do not know Ruby.
- [x] XML documentation is included with the NuGet packages.
- [x] FAKE dependency is in a separate group.
- [x] Each package gets its own description.
- [ ] Direct pushing to NuGet is not yet implemented, because an API key is needed. But it can be implemented in AppVeyor.
- [ ] The Travis CI build does not yet work, due to inadequate knowledge of its workings.

There are also these requirements of the build system:

- [x] Have a single target that compiles both for Mono and Core
- [x] Versioning information should be in a single location; a file that can be edited from the command line. My current workflow is to run `semver bump minor` (or patch, or major, or `semver special beta.2`)
- [ ] Have a way to override per-project properties in the resulting nugets (paket.template?)
- [ ] There must be a way to show what nuget package xml is generated, preferrably by prepending `DEBUG=true` to the command line to set the env var.
- [x] There must be a target to push all nugets to nuget.org, by taking the nuget key from the environment variables
- [x] There must be a check that there are no pending changes when the `release` target is run
- [x] The release target must run the tests
- [x] The release target must compile release notes and append them to the release notes of the nuget
- [x] The release target must first verify that the git-tag of the version-file is not taken, and then
- [x] tag the git release
- [x] Push the nugets
- [x] Push the git tag and git commit
- [ ] If any of the above steps fail, run the chain backwards as a compensatory action. Nothing must be pushed to github if the release fails for some reason.
- [ ] Finally; the build must not be flaky.


.version
```diff
@@ -0,0 +1 @@
+5.0.0
```
teo-tsirpanis: This file is the only source of version information. I could use the CI build scripts, but it will not work from local builds.

build.fsx
```diff
@@ -0,0 +1,87 @@
...
+// Build order
+"ApplyVersion"
+    ==> "Clean"
+    ==> "AssemblyInfo"
+    ==> "FixLogary"
+    ==> "Restore"
+    ==> "Build"
+    ==> "CleanBuildOutput"
+    ==> "Pack"
+"Build" ==> "Test"
```
teo-tsirpanis: The tests are not included in the default build pipeline, because some "deliberately failing tests" break the build.

global.json
```diff
@@ -1,5 +0,0 @@
-{
```
teo-tsirpanis: `global.json` is removed, asit's deprecated and succeeded by solution files.

paket.dependencies
```diff
@@ -1,18 +1,13 @@
-source https://nuget.org/api/v2
+source https://api.nuget.org/v3/index.json
```
teo-tsirpanis: The NuGet API v3 URL is used. This is mostly a cosmetic change.

Rakefile
```diff
@@ -36,7 +36,7 @@ task :quick_compile_netcore do
 end
 
 task :paket_bootstrap do
-  system 'tools/paket.bootstrapper.exe',
+  system 'tools/paket.bootstrapper.exe prerelease',
          clr_command: true unless File.exists? 'tools/paket.exe'
 end
```
teo-tsirpanis: The Rakefile is includud only for reference purposes.

Expecto.Tests/Tests.fs
```diff
@@ -19,12 +19,7 @@ module Dummy =
   [<Tests>]
   let testB() = TestLabel ("test B", TestList ([], Normal), Normal)
 
-  let thisAssemblyName =
-#if NETCOREAPP1_1
-    "Expecto.netcore.Tests"
-#else
-    "Expecto.Tests"
-#endif
+  let thisAssemblyName = "Expecto.Tests"
```
teo-tsirpanis: There is now one assembly name for .NET Framework and .NET Core projects.

tools/version.props.template
```diff
@@ -0,0 +1,14 @@
+<Project>
```
teo-tsirpanis: This file contains the common NuGet package metadata. It gets renamed to `version.props` during the build process.

tools/version.props.template
```diff
@@ -0,0 +1,14 @@
...
+  <PropertyGroup>
+    <Version>@Version</Version>
```
teo-tsirpanis: The `@Version` tag is replaced by the contents of the `.version` file during the build process.

tools/version.props.template
```diff
@@ -0,0 +1,14 @@
...
+    <Authors>Henrik Feldt, Logibit AB, formerly @mausch</Authors>
+    <Company>Steve Hatchett (Ported to .NET Standard by Theodore Tsirpanis)</Company>
+    <Copyright>Copyright © 2013-2017 SoftWx, Inc.</Copyright>
+    <PackageLicenseUrl>http://opensource.org/licenses/MIT</PackageLicenseUrl>
+    <PackageTags>testing fsharp assert expect</PackageTags>
+    <PackageIconUrl>https://raw.githubusercontent.com/haf/expecto/master/docs/expecto-logo-small.png</PackageIconUrl>
+    <RepositoryUrl>https://github.com/haf/expecto</RepositoryUrl>
+    <PackageProjectUrl>https://github.com/haf/expecto</PackageProjectUrl>
+    <Description>Expecto is a smooth test framework for F#, cloned from Fuchu with added functionality for making it easier to use.</Description>
```
teo-tsirpanis: This is the common description shared by all three packages. A specific description for each package will be written.

paket.dependencies
```diff
@@ -1,18 +1,13 @@
...
 nuget Argu
 nuget FsCheck
-nuget FSharp.Core
 nuget BenchmarkDotNet
-nuget BenchmarkDotNet.Toolchains.Roslyn
-nuget BenchmarkDotNet.Core
 nuget NuGet.CommandLine
 nuget Mono.Cecil >= 0.10-beta5
-
-group build
```
teo-tsirpanis: When I saw that FAKE was not used, I removed it, but I decided to use it as the build system. It will be in its separate group in the future.

haf:
Welcome to expecto, I'm glad to have you here as a contributor. Let me give a wish-list for replacing albacore:

 - Versioning information should be in a single location; a file that can be edited from the command line. My current workflow is to run `semver bump minor` (or patch, or major, or `semver special beta.2`)
 - Have a single target that compiles both for Mono and Core
 - Have a way to override per-project properties in the resulting nugets (paket.template?)
 - There must be a way to show what nuget package xml is generated, preferrably by prepending `DEBUG=true` to the command line to set the env var.
 - There must be a target to push all nugets to nuget.org, by taking the nuget key from the environment variables
 - There must be a check that there are no pending changes when the `release` target is run
 - The release target must run the tests
 - The release target must compile release notes and append them to the release notes of the nuget
 - The release target must first verify that the git-tag of the version-file is not taken, and then
 - tag the git release
 - Push the nugets
 - Push the git tag and git commit
 - If any of the above steps fail, run the chain backwards as a compensatory action. Nothing must be pushed to github if the release fails for some reason.
 - Finally; the build must not be flaky.

If these can be met, I'd welcome a change for the good of the community as well as for the equal maintainability level for me.

build.fsx
haf: We don't have deliberately failing tests for Expecto. We have some in the samples, but they should not be run on build.

Rakefile
haf: And if you reference paket elsewhere, please make it so that we lock it down to a version, so that we're not eternally on a prerelease version.

Expecto.FsCheck/Expecto.FsCheck.fsproj
```diff
@@ -1,105 +1,19 @@
...
-    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
-    <ProductVersion>8.0.30703</ProductVersion>
-    <SchemaVersion>2.0</SchemaVersion>
-    <ProjectGuid>{58507646-335a-460e-b283-77a2baf0d3a8}</ProjectGuid>
-    <OutputType>Library</OutputType>
-    <RootNamespace>Expecto.FsCheck</RootNamespace>
-    <AssemblyName>Expecto.FsCheck</AssemblyName>
-    <TargetFrameworkVersion>v4.6.1</TargetFrameworkVersion>
-    <Name>Expecto.FsCheck</Name>
+    <TargetFrameworks>net461; netstandard1.6</TargetFrameworks>
```
teo-tsirpanis: > Have a single target that compiles both for Mono and Core

The requirement is met.

isaacabraham:
Some of the other elements could be met by looking at the project scaffold samples I imagine.

build.fsx
teo-tsirpanis: I think there are [some in FsCheckTest.fs](https://github.com/haf/expecto/blob/master/Expecto.Tests/FsCheckTests.fs#L12).

The console says at the end:

[20:32:30 INF] EXPECTO! 166 tests run in 00:05:17.0965272162 passed, 2 ignored, **2 failed**, 0 errored. <Expecto>

build.cmd
```diff
@@ -1,13 +1,13 @@
 @echo off
 
-.paket\paket.bootstrapper.exe
+.paket\paket.bootstrapper.exe 5.0.0-beta008
```
teo-tsirpanis: > And if you reference paket elsewhere, please make it so that we lock it down to a version, so that we're not eternally on a prerelease version.

OK. It's done.

build.fsx
```diff
@@ -1,15 +1,26 @@
...
+#r "./packages/build/FSharp.Configuration/lib/net46/FSharp.Configuration.dll"
 
 open Fake
 open Fake.AssemblyInfoFile
 open Fake.Testing
+open FSharp.Configuration
 open System
 open System.IO
 
+type SemVer = YamlConfig<".semver">
```
teo-tsirpanis: > Versioning information should be in a single location; a file that can be edited from the command line. My current workflow is to run semver bump minor (or patch, or major, or semver special beta.2)

The requirement is met.

.semver
```diff
@@ -0,0 +1,6 @@
+---
```
teo-tsirpanis: The version info is again stored in the `.semver` file. I realized it was a YAML file, so it's easy to work with it.

haf:
I can add a nuget key to appveyor if we change to letting appveyor set the patch number.

teo-tsirpanis:
The patch number can be set by the `APPVEYOR_BUILD_NUMBER` environment variable, or by what the `.semver` file says, if we are on a local build.

> I can add a nuget key to appveyor

OK, you can push it, and I will merge it from upstream.

haf:
Here's an encrypted nuget key that's allowed to push:

```
environment:
  nuget_key:
    secure: vvcRRAmZ3qyDhkTae2A50uf75wycN+BA0QQ6OtTJxYAx2zPPi459C5CG/3Qr7z8J
```

TheAngryByrd:
> The release target must compile release notes and append them to the release notes of the nuget

Here's how I append the release notes to a nuget package 
```
let release = LoadReleaseNotes "RELEASE_NOTES.md"

Target "DotnetPack" (fun _ ->
    !! srcGlob
    |> Seq.iter (fun proj ->
        DotNetCli.Pack (fun c ->
            { c with
                Project = proj
                Configuration = "Release"
                OutputPath = IO.Directory.GetCurrentDirectory() @@ "dist"
                AdditionalArgs = 
                    [
                        sprintf "/p:PackageVersion=%s" release.NugetVersion
                        sprintf "/p:PackageReleaseNotes=\"%s\"" (String.Join("\n",release.Notes))
                    ]
            }) 
    )
)
```
https://github.com/TheAngryByrd/MiniScaffold/blob/master/Content/build.fsx#L131

teo-tsirpanis:
@TheAngryByrd, that's a nice idea. 😲

@haf, actually, why push the release NuGet packages from AppVeyor? This is not a procedure that should happen after each build. AppVeyor has [its own NuGet feed](https://www.appveyor.com/docs/nuget). I recommend that two targets should be created. The first one will be run after each build, and it will publish the packages to AppVeyor's feed. The second one will publish them to nuget.org, and it will be only run manually.

haf:
@teo-tsirpanis You decide. But I don't think commits on master should be delayed in their releases

teo-tsirpanis:
I got [a better idea](http://i3.kym-cdn.com/photos/images/facebook/001/217/711/afd.jpg_large). If the commit message begins with "Bumped Version to \<something>", it will push _both_ to nuget.org and AppVeyor, otherwise, only to AppVeyor. This means that the build script should not make a new commit, just a new tag. I need an API key for the AppVeyor feed.

> The release target must compile release notes and append them to the release notes of the nuget

This seems easy. But where are the release notes? The commit messages could be the release notes but I dont's think it's a good idea (they will be full with lines like "a small fix", "another small fix", "OK, this must now work" 😄).

haf:
The release notes could be all titles of all merged PRs, perhaps?

teo-tsirpanis:
> The release notes could be all titles of all merged PRs, perhaps?

That would be pretty hard and complicated. I think it's better to make a release notes file, just like many other projects.

RELEASENOTES.md
```diff
@@ -0,0 +1,3 @@
+### New in 5.0.0
```
haf: So, an alternative idea then, perhaps it would be useful to keep an empty release notes file and enforce it has contents on release? Then as the release finishes, the file is emptied again?

haf: Because we still need to push the git tag corresponding to the actual release. So it's not useful if Appveyor does the release and doesn't push the git tag. So if I instead build it locally, I can both manage the release notes and the tagging via the build script.

haf:
What do you mean with needing an API key?

RELEASENOTES.md
teo-tsirpanis: I think you might have misunderstood the release pipeline proposal.

1. You (only commits by haf will trigger it) want to release a new version.

2. You add at the _top_ of the RELEASENOTES.md file something like this:
``` markdown
### New in 6.0.0 (Released in 2017/12/14)

* Added support for .NET Core 3.5
```

3. You update the `.semver` file accordingly.

4. You make a commit that begins with the phrase `bumped version` (case-insensitive) and push it.

5. AppVeyor will see it, build the packages, push them to NuGet, make a Git tag, and push it too. It will not build the project _again_, after the tag is pushed.

The release notes are updated, _not_ replaced. I just did not add the previous release notes. It would be tedious.

teo-tsirpanis:
AppVeyor has its own NuGet repository, and it could be used for bleeding-edge packages.
An API key and feed URL can be found [here](https://ci.appveyor.com/nuget). They will be integrated just like nuget.org feeds and API keys.

build.fsx
haf: But that should not be run: https://github.com/haf/expecto/blob/master/Expecto.Tests/FsCheckTests.fs#L6 – there's no `Tests` attribute on it.

haf:
Travis is failing due to a bug in FAKE, it would seem:

```
Error: System.Exception: Unknown build environment: Travis

	  at FSI_0005.Build+shouldPushToGithub@80.Invoke (System.String message) [0x00001] in <6c6d15f34ab54879a3ec467e343672e4>:0 
```

teo-tsirpanis:
> Travis is failing due to a bug in FAKE, it would seem:

No, it's not a bug in FAKE.. [It's my fault, sorry :disappointed:.](https://github.com/haf/expecto/pull/175/files#diff-6074983490330a0aac1bd23c958232d7R80) I will fix it next day. But Travis fails either way.

tools/version.props
```diff
@@ -0,0 +1,12 @@
+<Project>
+  <PropertyGroup>
+    <Authors>Henrik Feldt, Logibit AB, formerly @mausch</Authors>
```
enricosada: If you are moving to msbuild 15 all projects..

if you move this file in `src` directory, and rename it `Directory.Build.props`, is automatically imported all msbuild project in each subdirectory.
So you can remove `<Import Project="..\..\tools\version.props" />`

That's [a new msbuild 15 feature](https://docs.microsoft.com/en-us/visualstudio/msbuild/what-s-new-in-msbuild-15-0).
Pratically each `fsproj` search recursively in the directory hierarchi for a `Directory.Build.props` (and `Directory.Build.targets` too) to import.

tests/Expecto.Tests/Expecto.Tests.fsproj
```diff
@@ -0,0 +1,31 @@
+<Project Sdk="FSharp.NET.Sdk;Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFrameworks>net461; netcoreapp1.1</TargetFrameworks>
+    <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
```
enricosada: `EnableDefaultCompileItems` is not needed, is already false

haf:
Hi guys,

I was wondering what the current status of this PR is? Do you need something from me?

haf:
I would prefer working with https://docs.travis-ci.com/user/encryption-keys/ in fact, because it command line all the way. This is about the TODO "Direct pushing to NuGet is not yet implemented, because an API key is needed. But it can be implemented in AppVeyor."

Travis is now failing with "/home/travis/build/haf/expecto/sample/Expecto.Sample/Expecto.Sample.fsproj.metaproj : error MSB4025: The project file could not be loaded. Could not find file '/home/travis/build/haf/expecto/sample/Expecto.Sample/Expecto.Sample.fsproj.metaproj'."

This seems to be a dotnet error or alternatively a scoping error, where we don't necessarily need to build the sample project for .net core (but why not, really...).

haf:
Ping

enricosada:
as a note, now `<TargetFrameworks>net461; netcoreapp1.1</TargetFrameworks>` is supported in FSAC based (vscode/vim/emacs), but atm the first target framework (`net461` in this example) is used for intellisense.

Work to ask user what want to use is in progress, final UX will be:

- or user is asked what use for this session, at first usage
- or read a configuration option to know what use (so is stored per repo/user)
  - plus a config to just use the first, like now, because usually is a sane default

teo-tsirpanis:
Well, this PR has been proven to be too much for me to successfuly handle. I made it to address two issues:

* Port everything to .NET Standard.
* Use a build script written in the same language as the project.

The first part is trivial, although the build script requires changes which I cannot do because I don't know Ruby. Moreover, VS cannot debug F# projects with the new project file format (but it's a functional language which does not need debugging; at least in theory 😛).

So, here we come to the second part. I tried to port the build system to FAKE, but it turned out that the requirements for the build system are not that simple. Given that it was the first time I was writing from scratch such sophisticated build system, I could not do that well. Things were complicated by two CI servers who did two different things, and lacking Ionide support.

Recently, I came accross ProjectScaffold. For [a project I am currently working on](https://teo-tsirpanis.github.io/Farkle/), it was a perfect skeleton (although I made many modifications on the top of it). One could copy from that concept. Excepto does not have a project website, so this makes things much simpler.

But I have personal projects to do too, and do not want to enter this hassle in the near future.

haf:
@teo-tsirpanis Yeah, you chewed on a few different things there. I think just chewing on .Net Core is enough for a PR, which you've done.

What's left of the Net Core port? I tried to compile it yesterday, but it's for v1.x of Core. If you can make it build for core 2.0 I can fix the Rakefile?

teo-tsirpanis:
> What's left of the Net Core port?

The libraries target .NET Standard 1.6 which can be run by .NET Core 2.0. If you mean the tests, they run on .NET Core 1.1. Why port them to 2.0? I guess it can run 1.1 apps too.

haf:
@teo-tsirpanis Could you extract the bits that work well for .Net Standard, perhaps? Then we could salvage part of this PR.

PR #180, opened 2017-06-16
Expect.throws should fail when exception is not raised

alex-bogomaz:
Hello, please review this PR.
It seems that now ```Expect.throws``` does not fail test when exception is not raised


haf:
Thank you!

PR #183, opened 2017-06-24
fix link to benchmarkdotnet.org

tomzo:


tomzo:
There is https://benchmarkdotnet.org/ but I get cert error when I browse it.

But previous link is broken certainly

haf:
Thanks!

PR #184, opened 2017-06-24
Update information about how to install Expecto Template

MNie:
Since Expecto.Template has stable version it could be installed without specifying a concrete version.

@haf


haf:
Thank you @MNie !

PR #189, opened 2017-08-04
add Func<Task> overloads to the C# api

0x53A:
Reason:

The following code compiles and runs, but does not what I expected it to:

```C#
        [FTests]
        public static Test blah =
            Runner.TestList("Test", new[]{ Runner.TestCase("TestCase", async () =>
            {
                await Task.FromException(new Exception("FAIL"));
            })});
```

image

It used the ``Action`` overload, so the exception is thrown on the threadpool, where no-one cares.

Now it _should_ prefer the ``Func<Task>`` overload and correctly fail the test.

Note: I can't build it locally, so let's see what the CI says.

### DIFF
--- a/Expecto.Tests.CSharp/Tests.cs
+++ b/Expecto.Tests.CSharp/Tests.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Threading.Tasks;
 using Expecto.CSharp;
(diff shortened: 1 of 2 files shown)


baronfel:
Good find, and thanks for the extension!  This looks good to me, @haf

Issue #193, opened 2017-08-28
TypeLoadException: 'GetFilters' method missing from 'BenchmarkConfig' in 'Expecto.BenchmarkDotNet


swlaschin:
The `IConfig` interface in BenchmarkDotNet (v0.10.9) has a method `GetFilters` that is not implemented in [Expecto.BenchmarkDotNet.fs (v5.0.1)](https://github.com/haf/expecto/blob/756863d42d55e0671b9559ebeef1c33c40450b22/Expecto.BenchmarkDotNet/Expecto.BenchmarkDotNet.fs#L40)

(`GetFilters` was added in [this commit on May 13 2017](https://github.com/dotnet/BenchmarkDotNet/commit/fe3032b2f84d62b5c298e46a72caaba2ac166904))

When you run the test assembly, you get the following error:
```
TypeLoadException: Method 'GetFilters' in type 'BenchmarkConfig' from assembly 'Expecto.BenchmarkDotNet, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' does not have an implementation.
```
## Solution

Add support for `GetFilters` with the following changes to `Expecto.BenchmarkDotNet.fs`:

```
open BenchmarkDotNet.Filters    // new

type BenchmarkConfig =
    { ...
      filters : IFilter list }  // new

interface IConfig with
    ...
    member x.GetFilters()      = x.filters :> seq<IFilter>  // new

let benchmarkConfig =
    ...
    filters = def.GetFilters() |> List.ofSeq   // new
    }
```

I can provide these changes in a PR if you like.

haf:
@swlaschin Yes, please PR those changes; it would be appreciated.

haf:
Merged as fixed in #193

realvictorprm:
When is the new nuget released? I cannot use the current release due to this problem :(

PR #196, opened 2017-09-08
#193

MNie:
Changes which are needed to Expecto.BenchmarkDotNet reported by @swlaschin.
I also edit nuget source (if its not a problem) based on a one commit from #175 which is not merged.

CC: @haf


haf:
Thank you @MNie and @swlaschin

swlaschin:
Thank you @MNie!

Issue #186, opened 2017-07-13
Expecto should fail if you use same test name twice


forki:

baronfel:
What criteria were you thinking here? Case Sensitivity?

anderssonjohan:
@haf Is there anything more to do here or does #198 (merged) close this?

AnthonyLloyd:
@anderssonjohan yes and 5.1.0 is released now with this in.

PR #198, opened 2017-09-13
Feature/duplicated names

MNie:
#186 

I don't know if it suits @forki requirements, also I don't know if I do a check for duplicates in a proper place so if someone could review changes it would be great :).

CC: @haf 

When we have duplicates:
image
When we don't have duplicates:
image

### DIFF
--- a/Expecto.Tests/Tests.fs
+++ b/Expecto.Tests/Tests.fs
@@ -567,7 +567,7 @@ let expecto =
         ) |> assertTestFails
       ]
 
-      testList "string matches pattern" [
+      testList "string matches pattern for isMatch" [
         testCase "pass" <| fun _ ->
           Expect.isMatch "{function:45}" "{function:(\\d+)}" "string matches passed pattern"
 
@@ -576,7 +576,7 @@ let expecto =
         ) |> assertTestFails
       ]
 
-      testList "string matches pattern" [
+      testList "string matches pattern for isNotMatch" [
         testCase "pass" <| fun _ ->
           Expect.isNotMatch "{function:45d}" "{function:(\\d+)}" "string not matches passed pattern"
 
@@ -585,7 +585,7 @@ let expecto =
         ) |> assertTestFails
       ]
 
-      testList "string matches regex" [
+      testList "string matches regex for isRegexMatch" [
         testCase "pass" <| fun _ ->
           let regex = Regex("{function:(\\d+)}")
           Expect.isRegexMatch "{function:45}" regex "string matches passed regex"
@@ -596,7 +596,7 @@ let expecto =
         ) |> assertTestFails
       ]
 
-      testList "string matches regex" [
+      testList "string matches regex for isNotRegexMatch" [
         testCase "pass" <| fun _ ->
           let regex = Regex("{function:(\\d+)}")
           Expect.isNotRegexMatch "{function:45d}" regex "string not matches passed regex"
@@ -644,33 +644,33 @@ let expecto =
       ]
 
       testList "list is empty" [
-        testCase "pass" <| fun _ ->
+        testCase "list pass" <| fun _ ->
           Expect.isEmpty [] "list is empty"
 
-        testCase "pass" <| fun _ ->
+        testCase "array pass" <| fun _ ->
           Expect.isEmpty [||] "list is empty"
 
-        testCase "fail" (fun _ ->
+        testCase "list fail" (fun _ ->
           Expect.isEmpty [5] "list is not empty"
         ) |> assertTestFails
 
-        testCase "fail" (fun _ ->
+        testCase "array fail" (fun _ ->
           Expect.isEmpty [|5|] "list is not empty"
         ) |> assertTestFails
       ]
 
       testList "list is non empty" [
-        testCase "pass" <| fun _ ->
+        testCase "list pass" <| fun _ ->
           Expect.isNonEmpty [5] "list is non empty"
 
-        testCase "pass" <| fun _ ->
+        testCase "array pass" <| fun _ ->
           Expect.isNonEmpty [|5|] "list is non empty"
 
-        testCase "fail" (fun _ ->
+        testCase "list fail" (fun _ ->
           Expect.isNonEmpty [] "list is empty"
         ) |> assertTestFails
 
-        testCase "fail" (fun _ ->
+        testCase "array fail" (fun _ ->
           Expect.isNonEmpty [||] "list is empty"
         ) |> assertTestFails
       ]
--- a/Expecto.Tests/Tests.fs
+++ b/Expecto.Tests/Tests.fs
@@ -41,9 +41,6 @@ let tests =
     testCase "basic" <| fun _ ->
       Expect.equal 4 (2+2) "2+2"
 
-    testCase "basic" <| fun _ ->
-      Expect.equal 4 (2+2) "2+2"
-
     test "using computation expression" {
       Expect.equal 4 (2+2) "2+2"
     }
@@ -74,7 +71,6 @@ let tests =
         Expect.equal "Test" "Tes2" "Failing - string with different content"
       } |> assertTestFailsWithMsgStarting "Failing - string with different content.\n          Expected string to equal:\n          \"Tes2\"\n              ↑\n          The string differs at index 3.\n          \"Test\"\n              ↑\n          String does not match at position 3. Expected char: '2', but got 't'."
     ]
-  ]
 
     testList "sumTestResults" [
       let sumTestResultsTests =
--- a/Expecto.Tests/Tests.fs
+++ b/Expecto.Tests/Tests.fs
@@ -983,8 +983,8 @@ open System.Threading
 let stress =
   testList "stress testing" [
 
-    let singleTest =
-      testList "hi" [
+    let singleTest name =
+      testList name [
         testCase "one" ignore
       ]
 
@@ -996,10 +996,10 @@ let stress =
         }
       ]
 
-    let deadlockTest() =
+    let deadlockTest(name) =
       let lockOne = new obj()
       let lockTwo = new obj()
-      testList "deadlock" [
+      testList name [
         testAsync "case A" {
           repeat100 (fun () ->
             lock lockOne (fun () ->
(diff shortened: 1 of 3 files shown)


haf:
I'm going to let @forki weigh in here.

So the change is to let `--debug` print duplicate test names? Sounds good.

Could you update the README as well, please?

MNie:
@haf Yes I could, I will do that later today.
`So the change is to let --debug print duplicate test names? Sounds good.` not only with `--debug` :)

forki:
can't we make it fail!?
Let ask differently: when is it a good idea to have two tests with same name? one fails - now where do I look for the error?

MNie:
@forki 
`can't we make it fail!?` I think can we.
`when is it a good idea to have two tests with same name? one fails - now where do I look for the error?`. You mean that the check shouldn't be on the `full name` of a spec but only on a `name`? So instead of treating as a different specs, specs like this:
```
testList "a" [
    test "1" { Expect.equal 4 4 "4"
]

testList "b" [
    test "1" { Expect.equal 4 4 "4"
]
```

It should treat them as an exact the same specs? -> throw an error?

forki:
No these are different tests. I'd only let them fail if the full name is
not unique

Am 14.09.2017 10:30 schrieb "Michał Niegrzybowski" <notifications@github.com

MNie:
@haf I don't know if `Writting tests` is a proper section for information about that test names should be unique across project. Maybe better would be some new section named `Possible problems` or something?

@forki So in fact You expect what is done right now? Fail if a full name isn't unique with an information what names which are not unique across project.

Expecto/Expecto.fs
```diff
@@ -1594,7 +1594,7 @@ module Tests =
         | None -> runEval config tests |> Async.RunSynchronously
         | Some _ -> runStress config tests |> Async.RunSynchronously
       else
-        printf "Found duplicated test names, these names are: %A" duplicates
+        printf "Error found duplicated test names, these names are: %A" duplicates
```
haf: Need to print with the logger, otherwise this line will intermingle with other expecto output.

Expecto.Tests/Tests.fs
```diff
@@ -644,33 +644,33 @@ let expecto =
       ]
 
       testList "list is empty" [
-        testCase "pass" <| fun _ ->
+        testCase "list pass" <| fun _ ->
```
haf: Why change these? It's about **globally** unique names. I don't mind the change though, but I'm curious.

Expecto.Tests/Tests.fs
```diff
@@ -1020,17 +1020,17 @@ let stress =
 
     let sequencedGroup() =
       testList "with other" [
-        singleTest
-        testSequencedGroup "stop deadlock" (deadlockTest())
-        singleTest
+        singleTest "hi"
+        testSequencedGroup "stop deadlock" (deadlockTest("deadlock"))
+        singleTest "another hi"
```
haf: Please name this differently.

Expecto.Tests/Tests.fs
```diff
@@ -1020,17 +1020,17 @@ let stress =
 
     let sequencedGroup() =
       testList "with other" [
-        singleTest
-        testSequencedGroup "stop deadlock" (deadlockTest())
-        singleTest
+        singleTest "hi"
```
haf: Please name this differently.

Expecto.Tests/Tests.fs
```diff
@@ -1020,17 +1020,17 @@ let stress =
...
+        testSequencedGroup "stop deadlock" (deadlockTest("deadlock"))
```
haf: Extra parens; remove inner parenthesis.

Expecto.Tests/Tests.fs
```diff
@@ -1020,17 +1020,17 @@ let stress =
...
+        singleTest "another hi"
       ]
 
     let twoSequencedGroups() =
       testList "with other" [
-        singleTest
-        testSequencedGroup "stop deadlock" (deadlockTest())
-        testSequencedGroup "stop deadlock other" (deadlockTest())
-        singleTest
+        singleTest "hi"
```
haf: Please name this differently.

Expecto.Tests/Tests.fs
```diff
@@ -1020,17 +1020,17 @@ let stress =
...
+        testSequencedGroup "stop deadlock" (deadlockTest("deadlock"))
+        testSequencedGroup "stop deadlock other" (deadlockTest("other deadlock"))
+        singleTest "another hi"
```
haf: Please name this differently.

Expecto.Tests/Tests.fs
```diff
@@ -1020,17 +1020,17 @@ let stress =
...
 
     let twoSequencedGroups() =
       testList "with other" [
-        singleTest
-        testSequencedGroup "stop deadlock" (deadlockTest())
-        testSequencedGroup "stop deadlock other" (deadlockTest())
-        singleTest
+        singleTest "hi"
+        testSequencedGroup "stop deadlock" (deadlockTest("deadlock"))
+        testSequencedGroup "stop deadlock other" (deadlockTest("other deadlock"))
```
haf: Extra parens; remove inner parenthesis.

Expecto.Tests/Tests.fs
```diff
@@ -1051,7 +1051,7 @@ let stress =
             stressMemoryLimit = 0.001
             printer = TestPrinters.silent
             verbosity = Logging.LogLevel.Fatal }
-      Expect.equal (runTests config singleTest &&& 4) 4 "memory"
+      Expect.equal (runTests config (singleTest "hi") &&& 4) 4 "memory"
```
haf: Please rename.

Expecto.Tests/Tests.fs
```diff
@@ -1074,7 +1074,7 @@ let stress =
               stressTimeout = TimeSpan.FromMilliseconds 10000.0
               printer = TestPrinters.silent
               verbosity = Logging.LogLevel.Fatal }
-        Expect.equal (runTests config (deadlockTest())) 8 "timeout"
+        Expect.equal (runTests config (deadlockTest("deadlock"))) 8 "timeout"
```
haf: Extra parens not needed.

Expecto.Tests/Tests.fs
```diff
@@ -1106,7 +1106,7 @@ let stress =
             stress = TimeSpan.FromMilliseconds 100.0 |> Some
             printer = TestPrinters.silent
             verbosity = Logging.LogLevel.Fatal }
-      Expect.equal (runTests config singleTest) 0 "one"
+      Expect.equal (runTests config (singleTest "hi")) 0 "one"
```
haf: Please rename.

Expecto.Tests/Tests.fs
```diff
@@ -1117,7 +1117,7 @@ let stress =
             stressMemoryLimit = 0.001
             printer = TestPrinters.silent
             verbosity = Logging.LogLevel.Fatal }
-      Expect.equal (runTests config singleTest) 4 "memory"
+      Expect.equal (runTests config (singleTest "hi")) 4 "memory"
```
haf: Please rename.

Expecto.Tests/Tests.fs
```diff
@@ -1139,6 +1139,6 @@ let stress =
             stressTimeout = TimeSpan.FromMilliseconds 10000.0
             printer = TestPrinters.silent
             verbosity = Logging.LogLevel.Fatal }
-      Expect.equal (runTests config (deadlockTest())) 0 "no deadlock"
+      Expect.equal (runTests config (deadlockTest("deadlock"))) 0 "no deadlock"
```
haf: Extra parens not needed.

Expecto/Expecto.fs
```diff
@@ -1566,6 +1566,17 @@ module Tests =
...
 
+  /// Prints out names of all tests for given test suite.
+  let duplicatedNames test =
+    test
+    |> Test.toTestCodeList
+    |> Seq.toList
+    |> List.map (fun t -> t.name)
+    |> List.groupBy id
+    |> List.choose ( function
+      | _, x::_::_ -> Some x
```
haf: Newlines after `->`

Expecto/Expecto.fs
```diff
@@ -1566,6 +1566,17 @@ module Tests =
     |> Test.toTestCodeList
     |> Seq.iter (fun t -> printfn "%s" t.name)
 
+  /// Prints out names of all tests for given test suite.
+  let duplicatedNames test =
+    test
+    |> Test.toTestCodeList
+    |> Seq.toList
+    |> List.map (fun t -> t.name)
```
haf: prefer `List.groupBy (fun t -> t.name)` and then below `Some x.name`

Expecto/Expecto.fs
```diff
@@ -1577,9 +1588,14 @@ module Tests =
       1
     else
       let tests = config.filter tests
-      match config.stress with
-      | None -> runEval config tests |> Async.RunSynchronously
-      | Some _ -> runStress config tests |> Async.RunSynchronously
+      let duplicates = duplicatedNames tests
+      if duplicates |> List.isEmpty then
```
haf: Don't pipe single parameters.

Expecto/Expecto.fs
```diff
@@ -1577,9 +1588,14 @@ module Tests =
...
+        match config.stress with
+        | None -> runEval config tests |> Async.RunSynchronously
+        | Some _ -> runStress config tests |> Async.RunSynchronously
+      else
+        printf "Error found duplicated test names, these names are: %A" duplicates
```
haf: Always print with the loggers. (Also, should have been `printfn` not `printf`)

README.md
```diff
@@ -233,7 +233,7 @@ Expecto supports the following test constructors:
 All of the above compile to a `Test` value that you can compose. For example,
 you can compose a `test` and a `testCaseAsync` in a `testList` which you wrap in
 `testSequenced` because all tests in the list use either `Expect.fasterThan` or
-they are using `Expecto.BenchmarkDotNet` for performance tests.
+they are using `Expecto.BenchmarkDotNet` for performance tests. You have to remember that the full name of test(s) should be unique across your test project.
```
haf: Linebreak at 80 chars

haf: Just make this line a new paragraph and we're done. You can embolden the text **the fully qualified names of the tests need to be unique across your test project**.

Expecto/Expecto.fs
haf: Also, spaces on either side of `::`, please.

Expecto.Tests/Tests.fs
MNie: `pass` -> `list pass` because there are checks  for an array and list and they have the same name I would split them to different test lists.
image

PR #207, opened 2017-11-22
Add Expecto.Flip.Expect module to .NET Standard version

inosik:
The `Expecto.Flip.Expect` module is missing in the .NET Standard version.


AnthonyLloyd:
Great Thanks

Issue #223, opened 2018-02-27
Request: config option to pick name for logging (not hardcoded to "Expecto")


rmunn:
Currently when [Expecto sets up logging](https://github.com/haf/expecto/blob/dd476c4066b0e472a682598af72fd7f2d4ad6e86/Expecto/Expecto.fs#L415), it uses the name "Expecto" (hardcoded) for the logger name, so that any log output from Expecto has `<Expecto>` printed at the end. If you run several instances of Expecto at once, it can become unclear which log message comes from which instance.

The use case that leads me to ask for this is https://github.com/TheAngryByrd/MiniScaffold/, which runs Expecto tests under three different frameworks (currently `net461`, `netcoreapp1.1`, and `netcoreapp2.0`) by default. It runs all three tests in parallel, which can sometimes result in output like the following:

```plaintext
[10:29:27 INF] EXPECTO? Running tests... <Expecto>
[10:29:28 INF] EXPECTO! 1 tests run in 00:00:00.0955849 – 1 passed, 0 ignored, 0 failed, 0 errored. ᕙ໒( ˵ ಠ ╭͜ʖ╮ ಠೃ ˵ )७ᕗ <Expecto>
watch : Exited
watch : Waiting for a file to change before restarting dotnet...
[10:29:28 INF] EXPECTO? Running tests... <Expecto>
[10:29:28 INF] EXPECTO? Running tests... <Expecto>
[10:29:28 INF] EXPECTO! 1 tests run in 00:00:00.1374021 – 1 passed, 0 ignored, 0 failed, 0 errored. ᕙ໒( ˵ ಠ ╭͜ʖ╮ ಠೃ ˵ )७ᕗ <Expecto>
[10:29:28 INF] EXPECTO! 1 tests run in 00:00:00.0931585 – 1 passed, 0 ignored, 0 failed, 0 errored. ᕙ໒( ˵ ಠ ╭͜ʖ╮ ಠೃ ˵ )७ᕗ <Expecto>
watch : Exited
watch : Waiting for a file to change before restarting dotnet...
watch : Exited
watch : Waiting for a file to change before restarting dotnet...
```

If one of the frameworks had had a test failure while the other two passed, it would be hard for me to tell which test had failed. Was it the `netcoreapp1.1` framework? The `net461` framework? I don't know.

[...]

PR #224, opened 2018-02-28
Allow specifying log name at command line

rmunn:
A simple way to implement #223 without too many changes to the code.

### DIFF
--- a/Expecto/Expecto.fs
+++ b/Expecto/Expecto.fs
@@ -412,7 +412,8 @@ module Impl =
   open Helpers
   open Mono.Cecil
 
-  let logger = Log.create "Expecto"
+  let mutable logger = Log.create "Expecto"
+  let setLogName name = logger <- Log.create name
 
   type TestResult =
     | Passed
@@ -830,6 +831,8 @@ module Impl =
       printer : TestPrinters
       /// Verbosity level (default: Info).
       verbosity : LogLevel
+      /// Process name to log under (default: "Expecto")
+      logName : string option
       /// Optional function used for finding source code location of test
       /// Defaults to empty source code.
       locate : TestCode -> SourceLocation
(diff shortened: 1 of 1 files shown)


Expecto/Expecto.fs
```diff
@@ -1611,6 +1618,9 @@ module Tests =
     Global.initialiseIfDefault
       { Global.defaultConfig with
           getLogger = fun name -> LiterateConsoleTarget(name, config.verbosity, consoleSemaphore = Global.semaphore()) :> Logger }
+    match config.logName with
```
haf: option.logName |> Option.iter setLogName

rmunn: Done.

AnthonyLloyd:
Thanks. I will hold this here until I get ok from downstream repos that they can release. Its breaking so should be a major version for all of us, will check.

rmunn:
I would think that adding a command-line parameter would be the equivalent of adding a new API, a minor version number bump rather than a major bump (e.g., from 5.1 to 5.2 instead of to 6.0) since it's backwards-compatible (any build scripts that relied on the old API aren't affected since they don't use the new command-line parameter anyway).

Is there any reason why adding a new command-line parameter would be backwards-incompatible, i.e. would cause older build scripts to break?

rmunn:
... Ah, but it's not *necessarily* backwards-compatible for anyone using the `ExpectoConfig` record type, because that type has grown an extra item. Anyone who uses `let config = { defaultConfig with verbosity = Something }` would still find it backwards-compatible at a source level, but not necessarily at a compiled-DLL level.

Yeah, if it *could* be non-backwards-compatible, them by SemVer rules, we need to bump the major version number.

haf:
Better to bump major if it's not definitely backwards compatible and get people into that flow.

AnthonyLloyd:
yep, its going to be 6.0.0

PR #235, opened 2018-03-22
Minor grammar fix

rmunn:
The pronoun "his" feels wrong in English for an impersonal object like a test; "its" is the right pronoun here.


haf:
@rmunn thanks!

Issue #142, opened 2017-04-16
request to display list name in console output


jackfoxy:
It would be useful to have list names displayed along with the simple statistics of each test list run when NOT using the --debug switch. There is really no way to visually tell which lists ran. Example of current state:

[14:37:45 INF] EXPECTO? Running tests...
[14:37:48 INF] EXPECTO! 8 tests run in 00:00:03.5117416 - 8 passed, 0 ignored, 0 failed, 0 errored. Success!
[14:37:48 INF] EXPECTO? Running tests...
[14:37:51 INF] EXPECTO! 9 tests run in 00:00:02.1252166 - 9 passed, 0 ignored, 0 failed, 0 errored. Success!
[14:37:51 INF] EXPECTO? Running tests...
[14:37:51 INF] EXPECTO! 4 tests run in 00:00:00.3081461 - 4 passed, 0 ignored, 0 failed, 0 errored. Success!
[14:37:51 INF] EXPECTO? Running tests...
[14:37:51 INF] EXPECTO! 5 tests run in 00:00:00.2538840 - 5 passed, 0 ignored, 0 failed, 0 errored. Success!
[14:37:51 INF] EXPECTO? Running tests...
[14:37:51 INF] EXPECTO! 5 tests run in 00:00:00.3103716 - 5 passed, 0 ignored, 0 failed, 0 errored. Success!
[14:37:51 INF] EXPECTO? Running tests...
[14:37:52 INF] EXPECTO! 5 tests run in 00:00:00.2989135 - 5 passed, 0 ignored, 0 failed, 0 errored. Success!
[14:37:52 INF] EXPECTO? Running tests...
[14:37:53 INF] EXPECTO! 2 tests run in 00:00:01.5431141 - 2 passed, 0 ignored, 0 failed, 0 errored. Success!
[14:37:53 INF] EXPECTO? Running tests...
[14:37:54 INF] EXPECTO! 4 tests run in 00:00:00.2392289 - 4 passed, 0 ignored, 0 failed, 0 errored. Success!
[14:37:54 INF] EXPECTO? Running tests...
[14:37:56 INF] EXPECTO! 2 tests run in 00:00:02.7312977 - 2 passed, 0 ignored, 0 failed, 0 errored. Success!
[14:37:56 INF] EXPECTO? Running tests...
[14:37:57 INF] EXPECTO! 5 tests run in 00:00:00.2992306 - 5 passed, 0 ignored, 0 failed, 0 errored. Success!
[...]

PR #236, opened 2018-03-31
default printer log with list name

jackfoxy:
https://github.com/haf/expecto/issues/142

Prints nearest common ancestor list name in default printer summary log
If only 1 test, prints name or list name
If no common ancestor for summary, prints "miscellaneous"

### DIFF
--- a/Expecto/Expecto.fs
+++ b/Expecto/Expecto.fs
@@ -647,9 +647,34 @@ module Impl =
                 "( ರ Ĺ̯ ರೃ )"
               else
                 ""
+          let commonAncestor (parentNames : string list) =
+            let rec loop ancestor (descendants : string list) =
+                match descendants with
+                | [] -> ancestor
+                | hd::tl when hd.StartsWith(ancestor)->
+                    loop ancestor tl
+                | _ ->
+                    if ancestor.Contains("/") then
+                        loop (ancestor.Substring(0, ancestor.LastIndexOf "/")) descendants
+                    else
+                        "miscellaneous"
+
+            match parentNames with
+            | [x] -> x
+            | hd::tl ->
+                loop hd tl
+            | _ -> "miscellaneous" //can't get here
+
           logger.logWithAck Info (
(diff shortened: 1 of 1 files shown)


Expecto/Expecto.fs
```diff
@@ -647,9 +647,34 @@ module Impl =
                 "( ರ Ĺ̯ ರೃ )"
               else
                 ""
+          let commonAncestor (parentNames : string list) =
+            let rec loop ancestor (descendants : string list) =
+                match descendants with
```
haf: 2 space indentation is used throughout.

jackfoxy: ...sorry. I'll redo.

Expecto/Expecto.fs
```diff
@@ -647,9 +647,34 @@ module Impl =
...
+            | [x] -> x
+            | hd::tl ->
+              loop hd tl
+            | _ -> "miscellaneous" //we can't get here
+
           logger.logWithAck Info (
-            eventX "EXPECTO! {total} tests run in {duration} – {passes} passed, {ignores} ignored, {failures} failed, {errors} errored. {spirit}"
+            eventX "EXPECTO! {total} tests run in {duration} for {name} – {passes} passed, {ignores} ignored, {failures} failed, {errors} errored. {spirit}"
             >> setField "total" (List.sumBy (fun (_,r) -> r.count) summary.results |> commaString)
+            >> setField "name" (summary.results
```
haf: I don't like logging that does parsing conditional on the logging level (at least if you do an eta-expansion making `level` explicit; do you think you could extract the name to a field above?

haf:
👍 Looking good

Expecto/Expecto.fs
jackfoxy: OK, I think I know what you are asking for (but then in a comment below you say "looking good"??)
I don't know if I will be able to change it today. Hopefully in a couple days.

jackfoxy:
I think this is what you had in mind. Good requirement. The code looks much cleaner.

haf:
Nice! LGTM from me from a code-review perspective. Over to @AnthonyLloyd for the merge :)

AnthonyLloyd:
Perfect thanks!

PR #244, opened 2018-05-06
Fixed flipped throwsT not passing exception type through

daniel-chambers:
The flipped version of `throwsT` doesn't pass through the generic exception type parameter. The type inference engine is setting it to `obj` instead.

I've duplicated the existing `throwsT` unit tests and used the flipped version inside them. Without the fix, they fail with the following:

```
[23:08:57 ERR] expecto/expectations/flipped throwsT/pass failed in 00:00:00.
Should throw null arg. Expected f to throw an exn of type System.Object, but one of type System.ArgumentNullException was thrown.
  D:\Code\expecto\Expecto\Expect.fs(43,1): Expecto.Expect.throwsT@43-1.Invoke(String x)
 <Expecto>
```

### DIFF
--- a/Expecto.Tests/Tests.fs
+++ b/Expecto.Tests/Tests.fs
@@ -508,6 +508,23 @@ let expecto =
 
       ]
 
+      testList "flipped throwsT" [
+
+        testCase "pass" <| fun _ ->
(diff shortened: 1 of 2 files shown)


AnthonyLloyd:
Perfect PR. Thank you!

PR #245, opened 2018-05-07
fix: all and allEqaul call enumeration more than once.

kleidemos:


AnthonyLloyd:
Perfect PR. Many thanks!

Issue #248, opened 2018-05-21
Bug: expecto exists without waiting for console on error


haf:
When using `logger.{info,debug,...}`, the facade doesn't wait. If you want it to wait, use infoWithBP or the like.

/cc @AnthonyLloyd

AnthonyLloyd:
I'll review the use of the logger. Thanks

AnthonyLloyd:
resolved with #258

Issue #230, opened 2018-03-10
Sniff console colour and adapt if it's black


haf:
<img width="1217" alt="screen shot 2018-03-10 at 15 55 45" src="https://user-images.githubusercontent.com/193115/37243546-9947d65a-247b-11e8-8f5a-faeb3d8243a6.png">

Could be pushed to the Facade in fact.

Issue #190, opened 2017-08-23
Easiest way to group test output together per test?


Mnuzz:
Not a bug per se, more of a question. Is there an option to ensure that the test output is only printed to the console when the test is completed? And all of the output for that single test is printed together? This would help my potentially odd use case, where I have written tests that should fail but do not. It is difficult to tell which tests failed when there are many.

haf:
Solving #141 would enable this scenario.

I'm not sure if you would like to give it a go?

The idea is that, when each Expect.MethodX call instead throws something typed, we can collect that output for the test and implement printers the e.g. group all output in a tree; at the end of the round, or printed with https://github.com/logary/logary/tree/master/src/targets/Logary.Zipkin as the test progresses

PR #258, opened 2018-06-19
console work

### LINKED ISSUE #256: Logging licence
The logary facade is pretty great, cause it adds no new dependencies to the project. However, given the new licence of logary itself (which as far as I know is the only way to do much with the logs), you are no longer able to use logary in combination with 

> any/all Microsoft web stacks

This. As far as I interpret it (though, I'm in no way a lawyer), means I would not be able to test anything that uses the microsoft webstack, while at the same time doing anything with the log output (which requires more orchestration than the facade provides). This means, users are either going to have to stay on an old version of Logary (and receive no updates), or write their own connectors to the facade that takes the logs somewhere else. Is it maybe time to start thinking about changing the logging solution?

AnthonyLloyd:
This PR aims to resolve #256 #249 #248 #230 #190 #161 #152 #147

- [x] Flushable logger
- [x] log using ANSI colour codes
- [x] log to StringBuilder
- [x] progress indicator e.g. n/m tests or stress 50% complete
- [x] flush once complete or from first error
- [x] pretty print of ligatures etc

### DIFF
--- a/Expecto/Expecto.fs
+++ b/Expecto/Expecto.fs
@@ -1027,8 +1027,7 @@ module Impl =
     evalTests config test
 
   let private flush (l:Logger) =
-    let d = l :?> IDisposable
-    if isNull d |> not then d.Dispose()
+    if l :? IFlushable then (l :?> IFlushable).Flush()
 
   /// Runs tests, returns error code
   let runEvalWithCancel (ct:CancellationToken) config test =
--- a/Expecto/Expecto.fs
+++ b/Expecto/Expecto.fs
@@ -394,11 +394,8 @@ module Impl =
   open Helpers
   open Mono.Cecil
 
-  let ansiOutputWriter = new ANSIOutputWriter()
   let mutable logger = Log.create "Expecto"
   let setLogName name = logger <- Log.create name
-  let flushLogger() =
-    ansiOutputWriter.Flush()
 
   type TestResult =
     | Passed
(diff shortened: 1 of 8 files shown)


adamchester:
Hey @AnthonyLloyd this is looking really good!

I'm curious why you are choosing to create a [new logger](https://github.com/haf/expecto/pull/258/files#diff-1d108374b84010eb70bf3a856bb172cfR944) instead of modifying the `LiterateConsoleTarget`(https://github.com/haf/expecto/pull/258/files#diff-1d108374b84010eb70bf3a856bb172cfL849)?

Is it possible (and easier?) to provide a new `outputWriter` that handles ANSI colour codes (only when the OS/terminal supports it)?

AnthonyLloyd:
I didn't really think of just creating a new outputWriter. I think because I needed to add Flush. I'll have a look at doing this as I integrate the progress indicator and flush on error. Thanks for the idea.

I tested the SetConsoleMode and it didn't have any effect. I think this is switched on by default now in Windows 10? Should possible consider for legacy versions.

adamchester:
Hey @AnthonyLloyd, as far as I can tell, it's not going to be just a "legacy" issue. Applications need to "opt in" to VT mode, and if you want to output ANSI sequences, you'll need to "opt in". I wrote about this topic in this issue because I actually experienced the ANSI escape codes being output on windows 10 (and VT mode was not enabled, so the just printed to the console).

See here for information directly from an MS person with knowledge of the situation: https://github.com/Microsoft/WSL/issues/1173#issuecomment-254250445

> There's a lot going on in here, so let me try to clarify: The windows console only supports VT sequences of ANY kind with the ConsoleMode set to enable Virtual Terminal support. See This documentation on SetConsoleMode. Specifically, you need to turn on the output mode flag ENABLE_VIRTUAL_TERMINAL_PROCESSING.

> By default, cmd.exe enables VT mode for itself. Now here's the confusion:

> In 10.0.10586 (or 1511 or TH2), it left this mode on when calling child applications. This meant that anything you launched from cmd would get the new behavior for free. Of course, we found that unsurprisingly broke compatibility, so we had to change that. So...
In 10.0.14393 (Anniversary update, or greater), cmd uses VT mode, but then reverts it for launching child processes. So if you wrote a console app that uses VT sequences after October 2015, then tried running it on an AU system, you'll find that the sequences don't work anymore. You'll have to manually turn on VT support with SetConsoleMode yourself. IMPORTANT cmd also wipes out the changes a child makes to the mode. So you can't call enable_vt.exe & myapp.exe and have the VT mode from enable_vt.exe pollute the mode of myapp.exe
To further confuse, bash.exe turns on VT mode itself, because obviously, it needs that.

AnthonyLloyd:
@adamchester Thanks, I'll add this in, best to be safe.

Expecto/Expecto.fs
```diff
@@ -1068,11 +1095,13 @@ module Impl =
...
-      let finishTimestamp =
-        lazy
+      let totalTicks =
         config.stress.Value.TotalSeconds * float Stopwatch.Frequency
         |> int64
-        |> (+) (Stopwatch.GetTimestamp())
+
+      let finishTime =
+        lazy
+        totalTicks |> (+) (Stopwatch.GetTimestamp())
```
haf: Remember that Stopwatch's ticks differs between operating systems and is not the same as DateTime/BCL ticks.

Expecto/Logging.fs
```diff
@@ -12,10 +12,12 @@
 ///
 /// Changes:
 ///     Changed namespace to Expecto.Logging and file name to Logging.fs - Anthony Lloyd - 11 Jun 2018
+///     Add IFlushable - Anthony Lloyd - 19 Jun 2018
```
haf: Why keep a changelog at the top? Feels a bit year 2000.

AnthonyLloyd: I read that it was a good thing to do when making changes to Apache code. To help when back porting the code. Could be got from git but it contains a bit more specific info.

PR #259, opened 2018-06-27
ExpectoConfig: add C# compatible With* methods / Logging: add Task based Interface

0x53A:
As the C# sample project shows, it is already possible without too much fuzz to define Tests in C#.

But without the F# ``with`` expression, it is actually pretty hard to update the config.

This adds .With* methods, and also adds a way for C# code to define a Printer.

### DIFF
--- a/Expecto/Expecto.fs
+++ b/Expecto/Expecto.fs
@@ -810,7 +810,42 @@ module Impl =
           do! innerPrinter.summary c s
           tcLog "testSuiteFinished" [
             "name", "ExpectoTestSuite" ] } }
+    static member fromInterface (i : ITestPrinter) =
+      { beforeRun   = i.BeforeRun >> Async.AwaitTask
+        beforeEach  = i.BeforeEach >> Async.AwaitTask
+        passed      = fun n d -> i.Passed(n, d) |> Async.AwaitTask
+        info        = i.Info >> Async.AwaitTask
+        ignored     = fun n m -> i.Ignored(n, m) |> Async.AwaitTask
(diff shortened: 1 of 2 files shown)


Expecto/Expecto.fs
```diff
@@ -810,7 +810,42 @@ module Impl =
           do! innerPrinter.summary c s
           tcLog "testSuiteFinished" [
             "name", "ExpectoTestSuite" ] } }
+    static member fromInterface (i : ITestPrinter) =
+      { beforeRun   = i.BeforeRun >> Async.AwaitTask
+        beforeEach  = i.BeforeEach >> Async.AwaitTask
+        passed      = fun n d -> i.Passed(n, d) |> Async.AwaitTask
```
haf: This makes the async run-once instead of run-many. Better to do `async { return i.BeforeRun }` or the like (delaying the construction of the `Task` value).

haf:
Minor fix @0x53A before merge — but I think we can merge this with that change.

AnthonyLloyd:
@0x53A @haf it still needs to be moved to CSharp namespace.

0x53A:
I'll fix the async issue - good catch, thank you.

Which parts should be moved to the CSharp namespace? Just the interface? Or should I also move the with* methods to extension methods and the CSharp namespace?

I'll do it tomorrow, because I currently don't have internet at home.

AnthonyLloyd:
@0x53A Everything that is in Expecto.fs. I think it can all be extention methods. Thanks

0x53A:
Ok the two should be done.

Can you please take a look again?

Thank you both.

AnthonyLloyd:
Thank you very much!

PR #263, opened 2018-07-06
[WIP] upgrade fake

AnthonyLloyd:


baronfel:
I _think_ the problem is that the `mono` executable isn't being invoked when the full-framework build of the tests is going (see here: https://github.com/haf/expecto/blob/fbdddff2c4077092d7fb8443376cdc04d0e4bdc7/build.fsx#L109). you may need to alter the `Process.execSimple` call to have `mono` be the fileName and the project exe be just another argument.

Issue #269, opened 2018-08-03
For Task based tests, exceptions are printed twice


0x53A:
![grafik](https://user-images.githubusercontent.com/4236651/43632498-a0066d3c-9706-11e8-8d11-62dd033e0718.png)

adamchester:
Hi @0x53A, there is only one exception there (although it is an aggregate exception, so it contains one or more inner exceptions too), and the output is more or less what ToString() on the exception would generate.

Does this help?

0x53A:
Hi @adamchester, yes, the output is correct.

I didn't write it (I should have maybe elaborated a bit more instead of just posting a screenshot ;) ), but my intention was to propose to unwrap AggreagateExceptions.

This is really common and annoying with Task based tests from c#.

What do you think of https://github.com/haf/expecto/pull/270?

adamchester:
My preference is not to hide the fact that it’s wrapped in an aggregate exception, but I can understand the argument for cleaning it up a little.

AnthonyLloyd:
I think this does simply for single exceptions. Will accept.

0x53A:
@adamchester for reference, the above screenshot was from this source code:

```c#
                Runner.FocusedTestCase("focused async", async () => {
                    throw new IOException("Hello World!");
                }),
```

adamchester:
:+1:

PR #270, opened 2018-08-03
unwrap AggregateExceptions with a single inner exception

0x53A:
fixes https://github.com/haf/expecto/issues/269

after:

grafik


AnthonyLloyd:
Thanks!

PR #272, opened 2018-08-28
Fix `throwsT` message on no exception.

otto-gebb:
In case when the function provided to `throwsT` threw no exception a misleading assert message was shown mentioning an `AssertException`.

The added test would fail with the following message:

> Should have failed with message containing: "Expected f to throw." but failed with "
> Should throw null arg. Expected f to throw an exn of type System.ArgumentNullException, but one of type Expecto.AssertException was thrown."


AnthonyLloyd:
Perfect, thanks!

PR #274, opened 2018-09-06
Create test results directory

JonCanning:


haf:
Doesn't that possibly throw?

JonCanning:
I think the exceptions for `CreateDirectory` are quite legitimate no?

haf:
Yeah.

Thank you

PR #282, opened 2018-09-26
junit report: add option to treat errors like failures 

0x53A:
because gitlab can't yet handle errors.

grafik

grafik

--------------

see https://gitlab.com/gitlab-org/gitlab-ce/blob/master/lib/gitlab/ci/parsers/junit.rb#L45-52
and issue https://gitlab.com/gitlab-org/gitlab-ce/issues/51087

This also means that skipped tests aren't surfaced anywhere and are treated as successfull. Which isn't great, but I guess they didn't fail ...

-------

I know you _just_ pushed a release yesterday, but for when do you plan the next?

Asking whether I should wait for official, or just build a nupkg myself.

Thank you!

### DIFF
--- a/Expecto.TestResults/CSharp.fs
+++ b/Expecto.TestResults/CSharp.fs
@@ -1,5 +1,6 @@
 namespace Expecto.CSharp
 
+open System.Runtime.InteropServices
 open System.Runtime.CompilerServices
 open Expecto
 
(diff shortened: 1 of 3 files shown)


AnthonyLloyd:
Thanks. I can release in the next couple of days. I have one other thing I want to put in.

PR #288, opened 2018-10-19
Document tests results are nunit v2

enricosada:
document that test results are nunit v2 format

upload test results in appveyor test tab (it's also useful as doc if others want to do that)

image


enricosada:
np.
I want to document that because I tried do it this week and lost a bit of time because I was trying to upload as schema v3, not v2 and appveyor error was criptic (just a 500)

PR #303, opened 2019-02-02
Args replacement for Argu

AnthonyLloyd:


haf:
Hmm, needs review? Did you just build a parser ;)

AnthonyLloyd:
Still v rough and I haven't tested it yet.

AnthonyLloyd:
@haf now fully functional as far as I can see.

Expecto.Tests/Tests.fs
```diff
@@ -181,6 +181,9 @@ let tests =
 
 [<Tests>]
 let expecto =
+
+  let testArgs i o = Expect.equal (Args.parseOptions options i) o "testArgs"
```
haf: Consider `sequence equal`

Expecto/Expecto.fs
```diff
@@ -154,8 +154,12 @@ module internal Helpers =
   let inline dispose (d:IDisposable) = d.Dispose()
   let inline addFst a b = a,b
   let inline addSnd b a = a,b
+  let inline fst3 (a,_,_) = a
   let inline commaString (i:int) = i.ToString("#,##0")
-
+  let inline tryParse (s:string) =
+    let mutable r = Unchecked.defaultof<_>
+    if (^a : (static member TryParse: string * ^a byref -> bool) (s, &r))
```
haf: Nice

Expecto/Expecto.fs
```diff
@@ -1613,6 +1626,83 @@ module Tests =
   /// The default configuration for Expecto.
   let defaultConfig = ExpectoConfig.defaultConfig
 
+  module Args =
+    open Microsoft.FSharp.Core
```
haf: `Microsoft.` is redundant

Expecto/Expecto.fs
```diff
@@ -1613,6 +1626,83 @@ module Tests =
...
+
+    type Parser<'a> = (string[] * int * int) -> Result<'a,string> * int
+
+    let parseOptions (options:(string * string * Parser<_>) list) (strings:string[]) =
+      let rec updateUnknown unknown last length =
+        if length=0 then unknown
+        else updateUnknown (strings.[last]::unknown) (last-1) (length-1)
+      let rec collect help unknown args paramCount i =
+        if i>=0 then
+          let si = strings.[i]
```
haf: `sCurr` or something easier to understand?

Expecto/Expecto.fs
```diff
@@ -1613,6 +1626,83 @@ module Tests =
...
+  module Args =
+    open Microsoft.FSharp.Core
+
+    type Parser<'a> = (string[] * int * int) -> Result<'a,string> * int
+
+    let parseOptions (options:(string * string * Parser<_>) list) (strings:string[]) =
+      let rec updateUnknown unknown last length =
+        if length=0 then unknown
+        else updateUnknown (strings.[last]::unknown) (last-1) (length-1)
+      let rec collect help unknown args paramCount i =
```
haf: `help` -> `isHelp`?

Expecto/Expecto.fs
```diff
@@ -1613,6 +1626,83 @@ module Tests =
...
+          if si = "--help" then
+            collect true (updateUnknown unknown (i+paramCount) paramCount) args 0 (i-1)
+          else
+            match List.tryFind (fst3 >> (=)si) options with
+            | Some(option,_,parser) ->
+              let arg, unknownCount = parser (strings,i+1,paramCount)
+              collect help
+                (updateUnknown unknown (i+paramCount) unknownCount)
+                (Result.mapError (fun i -> option + " " + i) arg::args) 0 (i-1)
+            | None -> collect help unknown args (paramCount+1) (i-1)
```
haf: A future PR could do fuzzy matching of input here

Expecto/Expecto.fs
```diff
@@ -1613,6 +1626,83 @@ module Tests =
...
+
+    let string case : Parser<_> =
+      fun (ss,i,l) ->
+        if l>0 then Ok(case ss.[i]), l-1
+        else Error "requires a parameter", 0
+
+    let list (parser:_->Parser<_>) case : Parser<_> =
+      fun (ss,i,l) ->
+        List.init l (fun j -> parser id (ss,i+j,1) |> fst)
+        |> Result.sequence
```
haf: `sequence >> map` = `traverse` if you have that

haf:
Looking good

Expecto.Tests/Tests.fs
AnthonyLloyd: It's comparing a `Result<args list,string list>` so a bit more tricky.

Expecto/Expecto.fs
AnthonyLloyd: currentArg

Expecto/Expecto.fs
AnthonyLloyd: Yes a bit like powershell

AnthonyLloyd: Powershell lets you shorten switches as long as they are unambiguous. This actually may not be a good idea for back compact reasons when new switches are added.

A possibly more useful one would be to suggest the correct switch based on a distance measure. Worth considering in the future. If there are enough good ideas I may pull Args out as a single file Args utility. Personally keen on where this could go as I use these quite a lot.

jzabroski: Got it.  I think the PowerShell folks got it wrong and they are starting to realize it.  With PowerShell 6, a group in Microsoft was looking into Query Auto Completion-like features to predict what you want to type for a command.  Sort of like next-gen IntelliSense for the command line. https://microsoft.github.io/prose/ Microsoft Program Synthesis using Examples SDK

Can you share how you use it a lot?

Simplest way to share would be to create a PowerShell Profile with a History logger extension so that you can record what you do in each session.  Then you can really data mine your use cases.

PR #312, opened 2019-03-29
Add Expect.hasLength to assert length of seq/list/array

drhumlen:


AnthonyLloyd:
Great. Pulling together a release next week, this will be included in that.

PR #317, opened 2019-04-04
Change default to 256, fixes #315

### LINKED ISSUE #315: Save our eyes; colours must have contrast by default
E.g. red on black has horrible contrast. https://www.viget.com/articles/color-contrast/

These are the de-facto colours now:

image

haf:


haf:
This is still an issue with 256 colours; they need tweaking, too

image

PR #346, opened 2019-09-22
Use DiffPlex to produce better diffs

drhumlen:
Hi! I've long been wishing for better diffs when comparing bigger objects/records or strings. I've searched around for a library to do better string diffing, and found a library called DiffPlex that does just that. I've posted some example screenshot below of the current diff rendering vs DiffPlex's. Imo it becomes a lot easier to see exactly what is changed in bigger object.

I've added some failing tests to the Sample project (not sure if it's the right place or not) to illustrate the new rendering.

This is still work in progress, but I rather show it off now before I put more time into it in case there's some reasons we can't do it this way. (Perfomance, compatability, preference..?). (Note also that I _temporarily_ removed net461 because I couldn't get net461 to run on my machine before my patience ran out. (I'm on MacOS and every installer from microsoft seems to be .exe files))

What do you think?

### Rendering examples:
Screenshot 2019-09-22 at 11 49 14
Screenshot 2019-09-22 at 11 48 08

-----------------------

Screenshot 2019-09-22 at 11 49 41
Screenshot 2019-09-22 at 11 47 43

-----------------------

Screenshot 2019-09-22 at 11 49 28
Screenshot 2019-09-22 at 11 47 53


haf:
@drhumlen Thank you for your suggestion and contribution! I hope you stick around and do more great stuff like this.

I really like the improved outputs, but I don't like taking on new dependencies. I see two alternatives;

- porting the algorithm here [which is possible under apache 2.0](https://github.com/mmanela/diffplex/blob/master/License.txt)
- creating a separate package, so that the diff is not as good by default, and making the diff algorithm configurable by adding a package

drhumlen:
> I really like the improved outputs, but I don't like taking on new dependencies.

Yeah, I understand that. It's always good to keep dependencies to a minimum to avoid being ... well _dependent_.

I was actually thinking about making the diffing function a user-configurable thing – so you can pass whichever differ function you want to use to Expecto. That way, you as a user can opt in on DiffPlex – or some other diffing library if you want to.

Porting it is also interesting. At first I thought this diffing algorithm was very complex, but it doesn't look too daunting looking at the code you linked. Perhaps just porting it over is the easiest and best result? 🤔

It's possible to do both also. Make the differ a configuration for the expecto user _and_ port over the algorithm as a "better default" (perhaps with the option to use the current differ).

drhumlen:
Screenshot 2019-09-22 at 21 35 42
Screenshot 2019-09-22 at 21 35 31

It's also possible to add a gutter with linenumber & color coding the change: blue=modified, green=whole line added, red=whole line deleted,  yellow="imaginary". Not sure if it's too much with numbers or not 🤔perhaps make it configurable? And of course not apply single-line diffs

haf:
> diffing function a user-configurable thing

This is not really necessary if we make it good by default; sometimes choice is a curse, too, and I've never seen any other testing library that lets the user change the diff function; or even seen anybody asking for that feature in the other libs I've used.

> Porting it is also interesting. At first I thought this diffing algorithm was very complex, but it doesn't look too daunting looking at the code you linked. Perhaps just porting it over is the easiest and best result? 🤔

Well, you'd have to port those 400-something LoC, but you could run the same tests that are already written in C# against your implementation, so you wouldn't have to port those. It would probably take an hour or four to port the code. It's probably a fun exercise and it would drastically improve the stability of this code base, compared to taking a dependency.

> It's also possible to add a ...

I'll let you decide how it should look for now; what you've done is much better than what we already have.

haf:
As for building with mono; you first [install mono](https://www.mono-project.com/download/stable/), and then do `source .env` in the repo.

drhumlen:
> Well, you'd have to port those 400-something LoC, but you could run the same tests that are already written in C# against your implementation, so you wouldn't have to port those. It would probably take an hour or four to port the code. It's probably a fun exercise and it would drastically improve the stability of this code base, compared to taking a dependency.

Yea, after reading some more through the library, it seems my initial gut-feeling was more right. There is significant work done by the DiffPlex, and it would be almost a little silly to rewrite all of that in F# – even if parts of it is trivial conversion.

What if we just grabbed the C# code directly and places it inside this repo instead of converting it to F#? I could go ahead and delete those (few) files we don't need to keep it small/simple. No reason to re-invent the wheel..? I guess it's possible to rewrite it in F#, but it's tedious work...

It's either that, or making improved diffs an addon with its own repo. But I agree with you, maybe that's making it _too_ flexible, and also a bit harder for a newcomer to have to add several packages just to get a nice diff.

haf:
> What if we just grabbed the C# code directly and places it inside this repo instead of converting it to F#? I could go ahead and delete those (few) files we don't need to keep it small/simple. No reason to re-invent the wheel..? I guess it's possible to rewrite it in F#, but it's tedious work...

Sure, let's go for this then; but then it would be a dependency named `Expecto.Plugins.Diffs` (or the like), and we have to design a way such that it gets used if it's available. So C# code in this repo; a copy (hi there @mmanela), which we can update from upsteam when needed. Including tests, which also should run as part of the CI pipeline. And no direct dependency on `Diffs` from the main `Expecto`; but only a replacement mechanism there. How does this sound?

mmanela:
Hi there, I am curious why the hesitancy of taking one dependency? You are free to copy the code, but it is often nicer for the community to take a lib dependency and the contribute changes and improvements back as you find things to enhance.

haf:
@mmanela I'm sure we'll repatriate code if we find improvements.

Why not a dependency:

I've taken "one dependency" before in Suave's source code, on FsPickler, believing it would be semver and schema compatible in minor releases, but it turned out it was not. The author thought this was fine.

The exact same happened for Argu in this lib, which caused us to build our own argument parser in this library. Logging; we're importing my Logary Facade lib, so that we can freeze its API, and Logary itself has a large test suite to avoid ABI incompatibilities.

In this case, looking at the list of releases https://www.nuget.org/packages/DiffPlex/ there's not a single version bump, which means that you've either kept the ABI/API breaking changes to zero the last 8 years, or there have been breaking changes that haven't been versioned as such. I haven't analysed the repository though.

Further, as this project is integrated into FAKE, Rider and VSCode, I think they deserve a stable API.

Taking a dependency also means we'll have to `#r "DiffPlex" as well as `#r "Expecto"` in any and all script files that use Expecto; this is why there's e.g. `Expecto.Hopac`; just so that we can have an abstract core that open for extension and closed for modification.

mmanela:
I understand and at least for diffplex, you are right I have kept the api very stable since there is not much need to change it

drhumlen:
> @drhumlen Thinking like above, perhaps we'll just have a 'sattelite' assembly and reference DiffPlex from it, like we do with Hopac; however, we'd still need the simple integration to happen once it's referenced.

@haf: Yeah, I'm trying to figure out the best way to do it now. On the top of my head, we could introduce a new function in Expect.fs:

```
let equal' diffPrinter (actual : 'a) (expected : 'a) message =
  ...
   if actual <> expected then
      failtestf "%s.\n%s" message (diffPrinter actual expected)

let equal = equal' printVerses  // <- the current behaviour/diff printer
```

and then let users import

```
open Expecto
open Expecto.DiffPlexDiffs

let expectEqual = Expect.equal' betterDiffPrinter
```

and then go on with their tests:

```
test "This is my test as an Expecto user" {
  expectEqual (1 + 1) 2 "One plus one is two"
}
```

instead of the usual `Expect.equal (1+1) 2 "One plus one is two"`

...

But that doesn't feel quite optimal if the rest of your tests starts with Expect._____

Alternatively, we could make Expect more of a class

```
type Expecto(diffPrinter) =
  member __.equal actual expected message =
    ....
  member __.isNone ......
```

and then let users do

open Expecto
open Expecto.DiffPlexDiffs

let Expect = Expecto(betterDiffPrinter)

test "This is my test" {
  Expect.equal (1+1) 2 "Business as usual here"
}

That also opens up rooms for inserting more user-configurable things into the Expect-module in the future.

But then again, I do like how the current Expect-module works without the need for instantiating a class. Hmm...

drhumlen:
A bit off-topic: When writing my own tests, I find that Expect.equal is the function is use by far the most. I also rather use more time to describe the test-case well instead of using the message-field – it's almost always redundant. I rather spend more time perfecting the name of helper functions and local values to make the test understandable.

...Which eventually led to me introduce a custom operator for assertions in my own test code base:

let (==>) actual expected = Expect.equal actual expected ""  // <- I don't need this message field

Then I can write my tests simply like so:

```
test "1+1 should be 2" {
   1 +1 ==> 2
}

test "When a user is created, it appears in the list of users with status Created." {
  let event: UserCreated = {Name = "Peter"}

  applyEvent [] event
  ==> [ {Name = "Peter"; Status = Created} ]
}
```
(We're doing event sourcing, so we have a lot of tests for our reducers (event appliers) & command handlers.)

I find this way of writing asserts lets me drop a lot of parentheses, while also being very visually guiding with the "arrow syntax". I initially had many tests written very nestedly with Expect.equal (......)  (..........) "", but I've slowly rewritten many tests to use the arrow and found that they become a more readable.

The reasons I'm mentioning it is because I think the Expect.equal function is by far the most useful one. You could do almost all the others things with it:   `user.IsVerified ==> true` instead of Expect.isTrue user "user must be verified".

I'm not sure if it's wise to pollute the namespace with a lot of symbolic operators, but perhaps _one_ is not too unreasonable – for example `==>` or some other suggestions if you have one? Maybe put it inside a module for symbols so it doesn't appear in scope when `open Expecto`, but enforce `open Expecto.OperatorSyntax`?

Just thougth I'd put it out there. Maybe this belongs in its own thread though.

drhumlen:
On-topic again: the easiest would be to just depend on DiffPlex (like i've done initially) and not bother with all the hassle of either porting code over to F# or confusing the end user with a plugin-like system.

...And if DiffPlex's API turns out to be unstable, we can always just go back to the old differ, or copy over the code that worked into this repo. Also, as long as we depend on a specific version of DiffPlex, I don't see why it should ever stop working? Also, @mmanela himself confirms that DiffPlex's api has been very stable and will probably continue to be since there's not much need to change it (ever)?

drhumlen:
Ok, I just went ahead trying out the Expecto.Hopac approach – making a seperate module and keeping Expecto (core) depedency-free.

I'm new to paket, and had to do a lot of wrestling with my IDE and Paket along the way, so some of the Paket-code/dependency tree maybe unnecessary/wrong. Please give feedback if there's something strange there.

Since the Colorization code is an internal/private module for Expecto, I have to pass a colorization function to the Expecto.Differ module's function – which looks very strange, but is is necessary to keep the color scheming consistent I think?

I'm still not quite done yet. String diffing doesn't use the provided differ function in every case, so that needs to be straightened out. But I'm awaiting feedback from you for now in case you've changed your mind on the dependency situation, or have a better solution :)

Expecto.Differ/Expecto.Differ.fsproj
```diff
@@ -0,0 +1,28 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <AssemblyName>Expecto.Differ</AssemblyName>
+        <PackageId>Expecto.Differ</PackageId>
+        <Version>8.12.0</Version>
+        <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
+        <GenerateDocumentationFile>true</GenerateDocumentationFile>
+        <!-- Optional: Declare that the Repository URL can be published to NuSpec -->
+        <PublishRepositoryUrl>true</PublishRepositoryUrl>
```
haf: I haven't seen this before? What's its effect?

Expecto.Differ/Library.fs
```diff
@@ -0,0 +1,45 @@
+module Expecto.Differ
+open System
+open DiffPlex.DiffBuilder
+open DiffPlex.DiffBuilder.Model
+
+let colorizedDiff (colorizeText: ConsoleColor -> string -> string) first second =
```
haf: We use `en-GB` in this repo, so `colourisedDiff`, if you please ;)

haf: Let's chat about the colourisation logic soon.

Expecto.Differ/Library.fs
```diff
@@ -0,0 +1,45 @@
...
+  let first, second =
+    match box first, box second with
+    | (:? string as f), (:? string as s) ->
+      string f, string s
+    | f, s ->
+      sprintf "%A" f, sprintf "%A" s
+
+  let differ = SideBySideDiffBuilder(DiffPlex.Differ())
```
haf: Is this object reusable? Hoist? Or is it stateful?

Expecto.Differ/Library.fs
```diff
@@ -0,0 +1,45 @@
...
+open DiffPlex.DiffBuilder
+open DiffPlex.DiffBuilder.Model
+
+let colorizedDiff (colorizeText: ConsoleColor -> string -> string) first second =
+  let first, second =
+    match box first, box second with
+    | (:? string as f), (:? string as s) ->
+      string f, string s
+    | f, s ->
+      sprintf "%A" f, sprintf "%A" s
```
haf: This will generate different indentation depending on the values of the fields in the objects. Of course, this is still an improvement, but a tree-walking algo that does consistent indentation would be preferred.

Expecto.Differ/Library.fs
```diff
@@ -0,0 +1,45 @@
...
+    match typ with
+    | ChangeType.Inserted -> colorizeText ConsoleColor.Green text
+    | ChangeType.Deleted -> colorizeText ConsoleColor.Red text
+    | ChangeType.Modified -> colorizeText ConsoleColor.Blue text
+    | ChangeType.Imaginary -> colorizeText ConsoleColor.Yellow text
+    | ChangeType.Unchanged | ChangeType.Imaginary | _ -> text
+
+  let colorizedDiff (lines: DiffPiece seq) =
+    lines
+    |> Seq.map (fun line ->
```
haf: split callback to own function, to avoid some indentation

Expecto.Differ/Library.fs
```diff
@@ -0,0 +1,45 @@
...
+      let styledLineNumber =
+        match line.Position |> Option.ofNullable with
```
haf: Single arguments are applied without `|>`.

Expecto.Differ/Library.fs
```diff
@@ -0,0 +1,45 @@
...
+        | Some num ->
+          (string num).PadLeft(2)
+          |> coloredText line.Type
+        | None -> "~~~"
```
haf: Could you give your reasoning behind the `-` and `~` chars you use to deliniate parts of the diff?

Expecto.Differ/Library.fs
```diff
@@ -0,0 +1,45 @@
...
+      if line.SubPieces.Count = 0 then
+        styledLineNumber + "  " + coloredText line.Type line.Text
+      else
+        let coloredPieces = line.SubPieces |> Seq.map (fun piece -> coloredText piece.Type piece.Text)
+        coloredPieces |> fun x -> styledLineNumber + "  " + String.Join("", x)
```
haf: Avoid extra lambda; you already have `colo[u]redPieces` as a constant above.

Expecto.Differ/Library.fs
```diff
@@ -0,0 +1,45 @@
...
+      )
+    |> fun x -> String.Join("\n", x)
+
+  sprintf "\n---------- Actual: --------------------\n%s\n---------- Expected: ------------------\n%s\n" (colorizedDiff diff.OldText.Lines) (colorizedDiff diff.NewText.Lines)
+
+let equals actual expected message =
```
haf: I see; so you're thinking we open `Expecto.Differ` and this symbol shadows `Expect.equal`? What about `sequenceEquals` and the ilk?

Expecto.Sample/Expecto.Sample.fs
```diff
@@ -36,6 +36,70 @@ let printingFsCheck =
       maxTest = 128
   }
 
+module ColorizationTests =
```
haf: Colourisation

Expecto.Sample/Expecto.Sample.fs
```diff
@@ -36,6 +36,70 @@ let printingFsCheck =
...
+    A: int
+    Nom: string
+    Weather: Weather
+    Tweets: int
+  }
+
+  let colorizationTests =
+    testList "Colorization tests. These should all fail to diplay the diff" [
+      test "Testing diff colorization" {
+        let personA = {Name = "Kesam"; Age = 30}
```
haf: Can we also have a test for a huge object where the fields differ a lot?

Expecto/Expect.fs
```diff
@@ -304,8 +304,13 @@ let stringStarts subject prefix message =
       "%s. Expected subject string to start with the prefix. Differs at position %i with subject '%c' and prefix '%c'.%s"
       message i s p (printVerses " prefix" prefix "subject" subject)
 
-/// Expects the two values to equal each other.
-let equal (actual : 'a) (expected : 'a) message =
+let colorizeString (color:ConsoleColor) text : string =
```
haf: Why is this public?

Expecto/Expect.fs
```diff
@@ -304,8 +304,13 @@ let stringStarts subject prefix message =
...
+  sprintf "%s%s%s"
+      (ANSIOutputWriter.colourText (ANSIOutputWriter.getColour()) color)
+      text
+      ANSIOutputWriter.colourReset
+
+let equalDiffer differ (actual : 'a) (expected : 'a) message =
```
haf: Let's chat about configuration

haf: Also, format params like `(myParam: int)`, not `(myParam:int)` and not `(myParam : int)`

haf:
BTW; DVar https://github.com/haf/DVar/blob/master/src/DVar.Tests/Program.fs — it's also present in Logary (and its Facade in this repo)

haf:
> I'm new to paket, and had to do a lot of wrestling with my IDE and Paket along the way, so some of the Paket-code/dependency tree maybe unnecessary/wrong. Please give feedback if there's something strange there.

Looks fine to me!

Expecto/Expect.fs
drhumlen: Yea, that's a helper. I'll make it private

Expecto.Sample/Expecto.Sample.fs
drhumlen: Absolutely. I just used tiny differences because those are the ones I'm usually struggling with spotting when the object is big and the diff colorization fails me.

I'll add a bigger object where many of the fields are wrong? Do you have an example in mind? Like {Address, Name, City, SecurityCard={Number, Expiration, CCV}, Pet=[{PetName=, PetSpecie}...]}  ish with many fields differentiating?

Expecto.Sample/Expecto.Sample.fs
drhumlen: I'll fix

Expecto.Differ/Library.fs
drhumlen: Actually I wasn't thinking about shadowing at all ... But shadowing is a possibility I guess 🤔

Expecto.Differ/Library.fs
drhumlen: Yea, I can experiment. I actually barely refactored yet at all – I just wanted to get things up and working quickly to see if it's worth finishing or not.

Expecto.Differ/Library.fs
drhumlen: Good question. I'm always unsure too when using non-functional first libraries. My guess is that the Builder has a side-effect and must be constructed every time.... But maybe the Differ() object is reusable?

Expecto.Differ/Library.fs
drhumlen: Alright, consistency is key :) 

Even though I'm a little confused by the choice of using "colour" etc. when the rest of the industry uses "color" – we're even using the ConsoleColor type here. But at the end of the day it doesn't matter of course as long as it's consistent. I'll fix!

Expecto.Differ/Expecto.Differ.fsproj
drhumlen: I just copied from Expecto.Halfo. I know almost nothing about these XML-files; i just really want them to work on my machine and for my IDE to recognize them :)  I'm basicly banging rocks together hoping for the depedencies to resolve.

drhumlen:
Thanks for the feedback. I'll refactor my code. It was never meant to be in a finished state; I just needed some feedback on the general direction before putting too much time into it.

I'm still the most concerned with: 1) How the user will "plugin" this and have it work without having to do anything weird. 2) How to pass the colourisation function to the new plugin. The way I've done it now works, but it's not too pretty 🤔

Expecto/Expect.fs
drhumlen: I agree. This is just haste-work ;) I'll fix formatting

drhumlen: Btw. I think the reason I formatted it this way was because the existing functions are formatted that way...?

drhumlen:
```
/// Expects the two values not to equal each other.
let notEqual (actual : 'a) (expected : 'a) message =
  if expected = actual then
    failtestf "%s. Actual value was equal to %A but had expected them to be non-equal."
        message actual
```

E.g. this one is formatted like this ^

Expecto.Differ/Library.fs
drhumlen: I can change it, but personally I prefer piping in this situation because then the most important thing, `line.Position`, comes first, and then the conversion from Nullable<->Option is more of a detail.

Expecto.Differ/Library.fs
drhumlen:
~~~ is now used for lines without a matching line on the other side. I guess a "gap" in the line numbering could work equally well. I wasn't really sure what it should look like, so this is just experimenting.

...I am leaning towards removing the whole line numbering/gutter rendering though. It's just too much imo – imo it's just as easy to read the diff without line numbers.

drhumlen:
Here I use a mutable variable to set the defaultDiffPrinter. I'm usually against mutable variables, but this gives us the un-intrusive plugin experience with only a single small mutable variable. Only catch is we have to remember to set the `defaultDiffPrinter` before the tests run. See the commit above^

drhumlen:
...Or, I guess the Expecto.Diff could just have the side-effect when running. But that's really dark magic: An import that modifies the program..?

drhumlen:
@haf :  Here's a big object diff
Screenshot 2019-09-26 at 23 53 08

drhumlen:
Feel free to checkout the branch and do the necessary steps to make this become mergeable. I'm not sure how to use DVar, or check if a module is available/loadable like you described.

Expecto.Differ/Library.fs
haf: `open Expecto.Flip` does exactly that, if you want an example ;)

Expecto.Sample/Expecto.Sample.fs
haf: No, it's enough if a field deep into a list of items has both long values in its fields and one of those fields differ. What I'm trying to trigger is that `%A` formats newlines if it gets to large fields.

drhumlen: Ah yeah i thought so. Imo it works satisfactory now, but perhaps you want to checkout the branch and play with a diff like that yourself?

drhumlen: Ah yeah i thought so. Imo it works satisfactory now, but perhaps you want to checkout the branch and play with a diff like that yourself?

drhumlen:
Hey. What will it take to get this merged now? The only thing that I feel is a bit "unsolved" now is the use of a variable to switch the differ function. It's possible to use a `ref` too, but that's essentially the same. Do you have any opinion on this? Is it acceptable?

Regarding tree-walking records, I think that's not really needed when you have a really good diff. You immediately see what the diff is, and the _whole_ diff, so ... is it needed? Tree-walking also only works for records afaik, so if you have an seq of records, or a record that contains a seq, then it won't help you anyway.

If I could wish for one thing (from F# in general?), it would be better pretty-printing. The default, %A, could've been so much nicer without the bloated semicolons and sometimes weird array indentation.

jzabroski:
This is pretty sweet and I'm going to steal this for my Xunit asserts.

haf:
@drhumlen No there's just the coordination of the config; we can merge but then I lock down master for releases until we have a nicer public reconfiguration API... Are you sure you don't want to take a stab at solving the DVar integration? :) There are examples of its usages in Logging.fs

haf:
🙌 @drhumlen  And thank you very much for the contribution :)

lydell:
Wow, this sounds amazing! I’m so happy I found this, and a bit sad I didn’t find it sooner. Would you mind explaining how to enable this? I can’t figure it out. 🙏 

@jzabroski Did you end up doing something similar for Xunit? I’d love to steal that for projects not using Expecto 😄

baronfel:
@lydell there's a code sample [here](https://github.com/haf/expecto/pull/346/commits/08f0cfc8cd250e456a890823645eb74a7fe46a66#diff-84c1e3ea21c2afef46062e5d2dcba123d9c20308d2a6882e7fa4da9dd75bef2aR29) in this PR demonstrating it, but you have to

* add the [Expecto.Diff](https://www.nuget.org/packages/Expecto.Diff/) package, and
* configure it using the code I linked above

lydell:
@baronfel Thanks you so much! 🥇 

Also, I tried using `Expecto.Diff.equals` in Xunit and it seems to work!

PR #347, opened 2019-09-22
update logary sample

smoothdeveloper:
The `Logary.Targets`  namespace seems required to get `LiterateConsole` in scope.


haf:
@smoothdeveloper yes that's correct; thank you!

Issue #348, opened 2019-10-04
Recursively print inner exceptions by default


cmeeren:
I propose that inner exceptions are printed by default. I am running an integration test using Hopac, and this is the test results I see:

![image](https://user-images.githubusercontent.com/7766733/66207265-a77c0200-e6b2-11e9-815d-629497d252a7.png)

This message says nothing about what is wrong. The inner exceptions eventually say stuff along the lines of "server not reachable", which is helpful, but I can't see that.

Let me know if there's a simple config adjustment for this (I couldn't find any config docs and was unable and unwilling to reverse-engineer it.)

haf:
Fixed with #349

cmeeren:
Wonderful, thanks! 🎉

cmeeren:
Would it be possible to release this on NuGet? 🙏

haf:
Yes, of course. Sorry for the wait. https://www.nuget.org/packages/Expecto/8.13.0 is the newest release out now!

cmeeren:
Thanks, but it doesn't seem to be fixed. I still get exactly the same result, just the "Exception raised by job" without any inner exceptions. (For good measure I deleted all bin/obj/.vs and restarted.)

haf:
cc @MNie

PR #349, opened 2019-10-06
Get inner message from exception

MNie:
Related to #348 . 
I have 2 questions:
1. All places that should be changed are covered here?
2. Checking if `InnerException` is right, or we prefer ` <> null`?

CC: @haf

### DIFF
--- a/Expecto/Expecto.fs
+++ b/Expecto/Expecto.fs
@@ -444,6 +444,14 @@ module Impl =
   let mutable logger = Log.create "Expecto"
   let setLogName name = logger <- Log.create name
 
+  let rec private exnWithInnerMsg (ex: exn) msg =
+    let currentMsg = 
+      msg + (sprintf "%s%s" Environment.NewLine (ex.ToString()))
+    if isNull ex.InnerException then
+      currentMsg
+    else
+      exnWithInnerMsg ex.InnerException currentMsg
+
   type TestResult =
     | Passed
     | Ignored of string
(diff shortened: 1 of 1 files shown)


Expecto/Expecto.fs
```diff
@@ -454,7 +454,18 @@ module Impl =
...
+          let currentMsg = 
+            msg + (sprintf "%s%s" Environment.NewLine (ex.ToString()))
+          if isNull ex.InnerException then
+            currentMsg
+          else
+            printInner ex.InnerException currentMsg
+
+        let exMsg = printInner e ""
+        
+        "Exception: " + exMsg
```
haf: Join together to `"Exception: " + printInner e ""`

Expecto/Expecto.fs
```diff
@@ -842,10 +853,18 @@ module Impl =
...
+              msg + (sprintf "%s%s" (ex.Message) Environment.NewLine)
+            if isNull ex.InnerException then
+              currentMsg
+            else
+              msgWithInner ex.InnerException currentMsg
+
           tcLog "testFailed" [
             "flowId", formatName n
             "name", formatName n
-            "message", e.Message
```
haf: Have you verified this works on TeamCity? Otherwise, I think we should just not change this location in this PR.

Expecto/Expecto.fs
```diff
@@ -454,7 +454,18 @@ module Impl =
       | Passed -> "Passed"
       | Ignored reason -> "Ignored: " + reason
       | Failed error -> "Failed: " + error
-      | Error e -> "Exception: " + e.ToString()
+      | Error e -> 
+        let rec printInner (ex: exn) msg =
```
haf: Wouldn't it be better to have this outside of the `TestResult` type to keep concerns separate?

haf:
I think we should focus on just solving the issue since changing it for the teamcity printer might have unintended consequences unless verified, such as having the message halfed due to the newline; TeamCity parses single lines into its messages.

haf:
Thank you for your PR @MNie ! This is what we need :)

Expecto/Expecto.fs
MNie: Done

Expecto/Expecto.fs
MNie: Agree, I move it to an `Impl` module and mark as a private.

Expecto/Expecto.fs
MNie: I couldn't do that for now, so I will remove changes from here for now :(

MNie:
@haf 
- changes for TC reverted;
- function moved to `Impl` module from `TestResult`;

haf:
@MNie Excellent, thank you!

PR #353, opened 2019-10-13
Extract helpers from large Expecto.fs file

haf:


haf:
This PR has no breaking nor API changes AFAIK

Issue #364, opened 2019-11-13
 Use dot as a separator instead of slash


cmeeren:
Originally posted in https://github.com/YoloDev/YoloDev.Expecto.TestSdk/issues/26, but I was redirected here.

When the VS test explorer renders the test tree, it seems to use dot as a separator (if grouping by namespace/class).

Consider the following tests:

```f#
[<Tests>]
let tests =
  testList "Domain" [
    testList "Username" [
      testList "create" [
        testCase "returns Error if input is empty" ignore
        testCase "returns Error if input length is > 100" ignore
        testCase "trims the input if Ok" ignore
      ]
    ]
    testList "CustomerNumber" [
      testList "create" [
        testCase "returns Error if input is empty" ignore
        testCase "returns Error if input is non-numeric" ignore
        testCase "returns Ok if the input is numeric" ignore
      ]
    ]
  ]
```

Test Explorer shows it like this:

![image](https://user-images.githubusercontent.com/7766733/68758044-59411380-060d-11ea-8add-572661847708.png)

All tests appear in a flat list. This makes it hard to read, hard to find relevant tests, and hard to select a related group of tests to run.

It would be much more useful if Test Explorer could show it as a tree. AFAIK this would only require using `.` instead of `/` as a separator in the hierarchy.

It can be worked around to some extent by adding dots to the end of each test list name:

```f#
[<Tests>]
let tests =
  testList "Domain." [
    testList "Username." [
      testList "create." [
        testCase "returns Error if input is empty" ignore
        testCase "returns Error if input length is > 100" ignore
        testCase "trims the input if Ok" ignore
      ]
    ]
    testList "CustomerNumber." [
      testList "create." [
        testCase "returns Error if input is empty" ignore
        testCase "returns Error if input is non-numeric" ignore
        testCase "returns Ok if the input is numeric" ignore
      ]
    ]
  ]
```

![image](https://user-images.githubusercontent.com/7766733/68758297-d66c8880-060d-11ea-8b01-30855002b581.png)

But it's not optimal: It requires changing the names, it still renders the slashes, and seems to only work for the last levels (e.g. `Domain./CustomerNumber` hasn't been split, and if there are single sibling test cases of nested tests lists, then they are not shown grouped in the same way).

If Expecto could support using `.` instead of `/` as a separator, I think that would give a much better experience in the VS test explorer. 🙂

Or if you have any other idea about how to improve this, that would be great!

---

[...]

PR #366, opened 2019-11-30
#364 split on

MNie:
Related Issue: #364 

I added a new option to config with a default value of `/`. 
@haf There is a single place that still uses a hardcoded '/' (Expecto.fs line 60). Should we care about that?


haf:
Yes, we need to care about that line, too, if we're to make this refactor. 

The alternative refactor is to represent test names as string arrays?

MNie:
@haf , yup, I will provide a pr tomorrow.

MNie:
@haf I change a label from `string` to `string list`. But there are a couple of questions.
- what with all those `helpers` like `testcasewithcancel` etc, Do we want to accept `string` there and then transform it to a list?
- what with filtering, we want to filter based on all names, on the first part of a name, last part?
- getting the last element in a list via `List.rev |> List.head` isn't the best solution, maybe you have some better idea? Array instead of a list? Or something else?

I didn't adjust tests, because I want to do a `checkpoint` if this is what we want to achieve?

Expecto.Tests/Tests.fs
```diff
@@ -131,7 +131,7 @@ let tests =
     testList "sumTestResults" [
       let sumTestResultsTests =
         let dummyTest = {
-          name = String.Empty
+          name = [String.Empty]
```
haf: Should be empty array

Expecto/Expecto.fs
```diff
@@ -170,6 +170,18 @@ module Impl =
     else
       exnWithInnerMsg ex.InnerException currentMsg
 
+  [<ReferenceEquality>]
```
haf: No ref eq

Expecto/Expecto.fs
```diff
@@ -291,19 +303,22 @@ module Impl =
     >> setField "errored" errored
     >> setField "erroredCount" (align erroredCount 0)
 
-  let createSummaryText (summary: TestRunSummary) =
-    createSummaryMessage summary Info
+  let createSummaryText splitter (summary: TestRunSummary) =
```
haf: It's a bit unclear what a splitter is?

Expecto/Expecto.fs
```diff
@@ -447,10 +454,10 @@ module Impl =
             let parentNames =
               summary.results
               |> List.map (fun (flatTest, _)  ->
-                if flatTest.name.Contains(splitSign) then
-                  flatTest.name.Substring(0, flatTest.name.LastIndexOf splitSign)
+                if (flatTest.name |> List.length) > 1 then
```
haf: Don't use the `|>` for single arguments

Expecto/Expecto.fs
```diff
@@ -447,10 +454,10 @@ module Impl =
...
+                  flatTest.name |> List.head
```
haf: Is the list backward?

Expecto/Expecto.fs
```diff
@@ -485,7 +495,7 @@ module Impl =
                   eventX "Maximum memory usage was {memory} KB and exceeded the limit set at {limit} KB.\nRunning tests:\n\t{timeout}"
                   >> setField "memory" (summary.maxMemory / 1024L |> int |> commaString)
                   >> setField "limit" (summary.memoryLimit / 1024L |> int |> commaString)
-                  >> setField "timeout" (summary.timedOut |> Seq.map (fun t -> t.name) |> String.concat "\n\t"))
+                  >> setField "timeout" (summary.timedOut |> Seq.map (fun t -> getName t.name) |> String.concat "\n\t"))
```
haf: Can't we put this on the type instead? I.e. `t` could have properties that do some of the transformations

Expecto/Expecto.fs
```diff
@@ -1612,10 +1623,10 @@ module Tests =
     | Fail_On_Focused_Tests -> fun o -> { o with failOnFocusedTests = true }
     | Debug -> fun o -> { o with verbosity = LogLevel.Debug }
     | Log_Name name -> fun o -> { o with logName = Some name }
-    | Filter hiera -> fun o -> {o with filter = Test.filter (fun s -> s.StartsWith hiera )}
-    | Filter_Test_List name ->  fun o -> {o with filter = Test.filter (fun s -> s |> getTestList o.splitter |> Array.exists(fun s -> s.Contains name )) }
-    | Filter_Test_Case name ->  fun o -> {o with filter = Test.filter (fun s -> s |> getTestCase o.splitter |> fun s -> s.Contains name )}
-    | Run tests -> fun o -> {o with filter = Test.filter (fun s -> tests |> List.exists ((=) s) )}
+    | Filter hiera -> fun o -> {o with filter = Test.filter (fun s -> s |> List.head |> fun z -> z.StartsWith hiera )}
```
haf: I don't really like to have thing that may crash inside this codebase, so please instead handle the distinct cases than use head.

Expecto/Expecto.fs
```diff
@@ -1612,10 +1623,10 @@ module Tests =
...
+    | Filter_Test_List name ->  fun o -> {o with filter = Test.filter (fun s -> s |> getTestList |> Array.exists(fun s -> s.Contains name )) }
+    | Filter_Test_Case name ->  fun o -> {o with filter = Test.filter (fun s -> s |> getTestCase |> fun s -> s.Contains name )}
+    | Run tests -> fun o -> {o with filter = Test.filter (fun s -> tests |> List.exists ((=) (s |> List.rev |> List.head)) )}
```
haf: Same

Expecto/Expecto.fs
```diff
@@ -1669,9 +1680,10 @@ module Tests =
         ArgsUsage (Args.usage commandName options, errors)
 
   /// Prints out names of all tests for given test suite.
-  let listTests test =
+  let listTests splitter test =
+    let split = Split.getSignAsString splitter
     Test.toTestCodeList test
-    |> Seq.iter (fun t -> printfn "%s" t.name)
+    |> Seq.iter (fun t -> printfn "%s" (String.Join(split, t.name)))
```
haf: Use `String.concat`

Expecto/Model.fs
```diff
@@ -88,7 +88,7 @@ type Test =
   /// A collection/list of tests.
   | TestList of tests:Test list * state:FocusState
   /// A labelling of a Test (list or test code).
-  | TestLabel of label:string * test:Test * state:FocusState
+  | TestLabel of label:string list * test:Test * state:FocusState
```
haf: Wouldn't it be possible to have a single label here? Because it's only the composition of these `Test` values that is the array; i.e. FlatTest would have an array, but a label shouldn't have to be an array

haf:
>   what with all those helpers like testcasewithcancel etc, Do we want to accept string there and then transform it to a list?

The behaviour from the CLI should be identical.

>    what with filtering, we want to filter based on all names, on the first part of a name, last part?

Same here. So you'd have to parse the string and/or join the test name. Perhaps use a value object and memoize the construction of the joined name?

>    getting the last element in a list via List.rev |> List.head isn't the best solution, maybe you have some better idea? Array instead of a list? Or something else?

Never use `head`, let's see how that turns out. Also, see if you can avoid pattern matching by only using the list when needed rather than for all labels?

haf:
BTW; this is a MAJOR version change. So I would like to have at least two other people who use Expecto testing this once tests are passing and it's stable.

haf:
Also thank you for refactoring this

Expecto/Model.fs
MNie: So simply keep here a name only relevant for a particular test (`test`), and when we log/get names then traverse the whole tree and build names like `groupA/subgroupAA/testA` or `groupA.subgroupA.testA`, right?

MNie: @haf upup :)?

haf: Yes exactly:)

haf: Yes exactly:)

Expecto/Expecto.fs
MNie: Done

Expecto/Expecto.fs
MNie: Removed

Expecto/Expecto.fs
MNie: Removed

Expecto/Expecto.fs
MNie: I changed the name to `joinBy` since the parameter means a sign by which we would join names of a test/test case/test list etc.

Expecto/Expecto.fs
MNie: Removed

Expecto.Tests/Tests.fs
MNie: After refactoring String.Empty will still be here.

Expecto/Expecto.fs
MNie: Agree, removed.

Expecto/Expecto.fs
MNie: Same (Agree, removed.).

Expecto/Expecto.fs
MNie: Done

Expecto/Model.fs
MNie: Done

Expecto.Tests/Tests.fs
```diff
@@ -611,27 +613,27 @@ let expecto =
           let count = ref 0
           let test = dummy (fun () -> incr count)
           runTestsWithCLIArgs [No_Spinner] [|"--filter"; "c/f"|] test ==? 0
-          !count ==? 2
 
         yield testCase "run with filter test case" <| fun _ ->
           let count = ref 0
           let test = dummy (fun () -> incr count)
           runTestsWithCLIArgs [No_Spinner] [|"--filter-test-case"; "a"|] test ==? 0
-          !count ==? 2
```
MNie: I have to remove those checks, cause `fun () -> incr count` is invoked before filtering. Perhaps because we have to get the name of a test. Some tips on how to keep the shape of those tests as they are before changes?

@haf

MNie:
@haf I fix'd a couple of things. If you would have a couple of minutes to take a look and give some feedback it would be great! ;)

Expecto.Tests/Tests.fs
haf: What changed in the implementation in this PR that makes increment be called before filtering?

Expecto/Expecto.fs
```diff
@@ -108,18 +110,18 @@ module Test =
...
     | TestList (testList, state) -> TestList (List.map (replaceTestCode f) testList, state)
     | TestLabel (label, test, state) -> TestLabel (label, replaceTestCode f test, state)
     | Sequenced (sequenced,test) -> Sequenced (sequenced,replaceTestCode f test)
 
   /// Filter tests by name.
-  let filter pred =
+  let filter joiner pred =
     toTestCodeList
-    >> List.filter (fun t -> pred t.name)
+    >> List.filter (fun t -> pred (t.name))
```
haf: I don't think we needed to change this?

Expecto/Expecto.fs
```diff
@@ -108,18 +110,18 @@ module Test =
...
     >> List.map (fun t ->
-        let test = TestLabel (t.name, TestCase (t.test, t.state), t.state)
+        let test = TestLabel ((t.fullName joiner), TestCase (t.test, t.state), t.state)
```
haf: no need for parens here either

Expecto/Expecto.fs
```diff
@@ -170,6 +172,23 @@ module Impl =
     else
       exnWithInnerMsg ex.InnerException currentMsg
 
+  type JoinBy =
+    | Dot
+    | Slash
+    member x.getSign () =
```
haf: I think we can make this

```fsharp
member x.sign =
  match x with Dot -> '.' | _ -> '/'
```

because evaluating it is extremely cheap

Expecto/Expecto.fs
```diff
@@ -170,6 +172,23 @@ module Impl =
...
+      match x with
+      | Dot -> '.'
+      | Slash -> '/'
+
+    member x.getSignAsString () =
```
haf: ...and `member x.signString`, for example

Expecto/Expecto.fs
```diff
@@ -259,10 +278,10 @@ module Impl =
       (if List.isEmpty x.timedOut then 0 else 8)
     member x.successful = x.errorCode = 0
 
-  let createSummaryMessage (summary: TestRunSummary) =
+  let createSummaryMessage joinBy (summary: TestRunSummary) =
     let handleLineBreaks (elements:(FlatTest*TestSummary) seq) =
         elements
-        |> Seq.map (fun (n,_) -> "\n\t" + n.name)
+        |> Seq.map (fun (n,_) -> "\n\t" + String.Join(joinBy, n.name))
```
haf: shouldn't we use the above mentioned properties?

Expecto.Tests/Tests.fs
MNie: @haf I don't see which change result in that, unfortunately :( Maybe if you could take a look you would spot a problem or have some hint?

teo-tsirpanis:
A better idea would have been to split the test names with dots, by default (for better Visual Studio experience and because of major release), and expose a `--join-by-slash` option to revert to the old behavior.

The way `--join-by` is formatted might mislead the user into thinking that _any string_ can be used to split the test names (flexible, but there is no reason to me that liberal with regards to configuration).

MNie:
@teo-tsirpanis I could change it to `.` as a default. Regarding the `--join-by` vs `--join-by-slash` from my point of view they are "equal". So if @haf is okay with that parameter name change I could also do that :)

haf:
Yes I'm fine with that suggestion, to change it to dot and add the 'revert to slash' flag. I wanted a second pair of eyes on this, so thanks @teo-tsirpanis .

Could you do this and rebase as well and we can get this merged into the upcoming major bump?

MNie:
@haf Yes, sure. I will do that today/over the weekend.

haf:
Thank you @MNie ! Great job!

Issue #373, opened 2020-03-24
Report test execution time


teo-tsirpanis:
My project has fifty tests. I observe the test runner to take some more time at one of them, but I have no idea which of them is the culprit.

Is there any way to view that? If not, a good implementation idea would have been to display execution time in the test summary, or even better create XML test reports like NUnit.

haf:
You could enable the xml reporting?

teo-tsirpanis:
Oh, I didn't know at all about that. I just saw the `Expecto.TestResults` NuGet package.

Thanks. Keep up the good work!

PR #374, opened 2020-03-24
Add command-line option to generate XML test report.

teo-tsirpanis:
This PR:

* Adds two command-line options `--nunit-summary` and `--junit-summary` that accept a string parameter and export a test summary in a respectively formatted XML file. It avoids incidents like #373 by making the feature more ~~accessible~~ known.
* __Removes the `Expecto.TestResults` project and moves its content to the main `Expecto` project.__ Usually, developers create a different package for the same library when it depends on something the main package shouldn't. Unlike `Expecto.FsCheck` or `Expecto.Hopac`, `Expecto.TestResults` has no such special dependencies and therefore should have been part of Expecto proper. This change allows for the former bullet point to be easily implemented.
* Splits the huge `Expecto.fs` file into three smaller ones: `Expecto.Impl.fs` which contains the `Impl` module, `Test.fs` which contains the `FlatTest` type and the `Test` module, and keeps the rest to `Expecto.fs`. Such a big file was hard-to-maintain, and splitting it was also necessary to implement the first bullet point.

The removal of the `Expecto.TestResults` assembly might be a breaking change, but I believe its introduction is totally justifiable, given how easy it will become to generate test reports without code changes.


README.md
```diff
@@ -134,7 +133,8 @@ and the output won't be interlaced due to concurrency.
 
 ### TestResults file
 
-    nuget Expecto.TestResults
+<s>
```
haf: What's this :)?

haf:
Well, it all looks great, like always when you're hacking away.

haf:
Did you manage to keep all public surface areas the same?

README.md
teo-tsirpanis: [A strikethrough](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/s). To indicate that the existing way of generating XML test results has changed.

teo-tsirpanis:
> Did you manage to keep all public surface areas the same?

Yes, I did. The only thing I removed was a misspelled identifier named `depricated` (had no reason to exist, and it had already been marked as obsolete).

Given that this change is going to be pushed for a major version bump, I am going to entirely remove the existing API to generate test results; it is already exposed via the more uniform `CLIArgs`.

haf:
> I am going to entirely remove the existing API to generate test results

Ok, so this is WIP still?

teo-tsirpanis:
Yes.

teo-tsirpanis:
It's ready.

haf:
@teo-tsirpanis thank you! I'll try release tonight

teo-tsirpanis:
Good, but a new _major_ version? That fast? I think a minor version increase (8.14.0) is more appropriate.

And don't forget to mark the `Expecto.TestResults` package as deprecated.

haf:
Sure, I have no problem delaying, but part of the contract with IDEs and users is that we don't break even the slightest thing in between minors :)

PR #375, opened 2020-03-25
Fix #328 and modernize the build system.

### LINKED ISSUE #328: Warning on config record: `parallel` is reserved for future use by F#
Using the `parallel` setting in a config results in a compiler warning:

    warning FS0046: The identifier 'parallel' is reserved for future use by F#

It's annoying to have a warning, but in future, I guess it could break.

I suppose it should be changed?

teo-tsirpanis:
This PR fixes #328 by renaming `ExpectoConfig.parallel` to `runInParallel` and modernizes the build system by upgrading CI images and integrating more with newer .NET Core SDK (like local FAKE and Paket tools and auto-generated AssemblyInfo).

_It is still a work in progress._


paket.dependencies
```diff
@@ -16,7 +15,7 @@ group Build
   storage none
   framework auto-detect
 
-  nuget FSharp.Core ~> 4.6.0
+  nuget FSharp.Core
```
teo-tsirpanis: There was no reason to pin the version of a development dependency. And it prevented Paket from getting the latest FAKE libraries.

NuGet.props
```diff
@@ -0,0 +1,24 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
+    <Description>Advanced testing library for F#</Description>
+    <Authors>Anthony Lloyd, Henrik Feldt and contributors</Authors>
+    <Owners>Anthony Lloyd &amp; Henrik Feldt (cloned from Fuchu by @mausch)</Owners>
+    <Copyright>Copyright 2019</Copyright>
+    <PackageRequireLicenseAcceptance>true</PackageRequireLicenseAcceptance>
```
teo-tsirpanis: Is there any reason for this one?

Expecto isn't using any restrictive or unusual license to merit requiring it to be approved.

teo-tsirpanis:
OK, I finished.

I rewrote major parts of the build script to rely more on .NET Core's project properties. These metadata were moved to a file called `NuGet.props` that is imported from the five library projects.

The .NET Core SDK was updated to 3.1.200 and FAKE and Paket were updated as well and are run from local tools.

I have a question about Travis: is it _really_ necessary to test against so many Mono versions? Four runs take a very long time, while we could only test against the latest Mono version, on Linux and on macOS.

NuGet.props
haf: Not really any good reason, no.

appveyor.yml
```diff
@@ -1,10 +1,6 @@
 image:
-  - Visual Studio 2015
-  - Visual Studio 2017
-
+  - Visual Studio 2019
```
haf: I'm happy limiting the number of monos this runs on to latest in travis.yaml.

teo-tsirpanis:
Done.

haf:
Great :)

What's your opinion on #366 ?  I'm thinking of merging that as well into this major release.

Issue #378, opened 2020-04-05
Filtering seems not to work


blair55:
I cannot replicate the documented filtering behaviour when running a test suite locally with Expecto 9.0.

I have taken the example below from the filtering unit test
https://github.com/haf/expecto/blob/498a4d2c5a675ff068a842b70bdfe271f374ead5/Expecto.Tests/Tests.fs#L572

Main.fs
```
  [<EntryPoint>]
  let main argv =
    printfn "%A" argv
    let fn = ignore
    TestList (
      [
        testCase "a" fn
        testCase "a_x" fn
        testCase "b" fn
        testList "c" [
          testCase "d" fn
          testCase "e" fn
          testList "f" [
            testCase "g" fn
            testCase "h" fn
          ]
        ]
      ], Normal)
    |> runTestsWithCLIArgs [] argv
```
Terminal
```
→ dotnet run -p tests --filter c
[|"--filter"; "c"|]                     
[22:46:59 INF] EXPECTO? Running tests... <Expecto>
[22:46:59 INF] EXPECTO! 7 tests run in 00:00:00.1037682 for miscellaneous – 7 passed, 0 ignored, 0 failed, 0 errored. Success! <Expecto>
```

All tests are run despite the existence of the `filter` arg.

PR #380, opened 2020-04-10
#378 pass filtered tests instead of not filtered.

MNie:
Fixes: #378

The wrong parameter was passed in Expecto.fs lines 523/524. The changes were introduce in this [commit](https://github.com/haf/expecto/commit/b168ca3e186b7c7bf5e4ac36fc99ec558aee2126) because of a rename from `tests` to `fTests` :(. 
 
CC: @haf


haf:
Thank you 😊

PR #383, opened 2020-04-14
Remove `My_Spirit_Is_Weak` since it was marked as `Deprecated`

MNie:
`/// Depricated. Will be removed on next major release.` -> So maybe release 10.0.0 is a nice candidate for that :)?

CC: @haf


haf:
Yeah, sure, but I'm not going to merge that in quite a while, since we just release a major :)

ratsclub:
Closed in favor of #449.

Issue #381, opened 2020-04-13
Benchmark 0.12.1 not working


Swoorup:
I tested out a sample benchmark as posted in the readme with installation of Expecto.benchmark package. It is failing with the following result: 

```
[06:28:21 ERR] performance tests.three serialisers errored in 00:00:00.0190000 <Expecto>
System.MissingMethodException: Method not found: 'System.Text.Encoding BenchmarkDotNet.Configs.IConfig.get_Encoding()'.
   at Benchmark.tests@42-1.Invoke(Unit unitVar)
   at Expecto.Impl.execTestAsync@566-1.Invoke(Unit unitVar)
   at Microsoft.FSharp.Control.AsyncPrimitives.CallThenInvoke[T,TResult](AsyncActivation`1 ctxt, TResult result1, FSharpFunc`2 part2) in F:\workspace\_work\1\s\src\fsharp\FSharp.Core\async.fs:line 398
```

MNie:
@Swoorup could you please give information, what packages (versions) are you using?
I try with the newest expecto (9.0.0) and expecto.benchmarkdotnet (9.0.0). It seems to work just fine.

![image](https://user-images.githubusercontent.com/7689103/79189400-edfca700-7e21-11ea-9794-207ae966a8f4.png)

Here is code I use to check the issue:

fsproj:
```fsharp
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp3.1</TargetFramework>
    <GenerateProgramFile>false</GenerateProgramFile>
  </PropertyGroup>

  <ItemGroup>
    <Compile Include="Main.fs" />
  </ItemGroup>

  <ItemGroup>
    <PackageReference Include="Expecto" Version="9.*" />
    <PackageReference Include="FSharp.Core" Version="4.*" />
    <PackageReference Include="Expecto.BenchmarkDotNet" Version="9.*" />
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.*" />
  </ItemGroup>
</Project>
```

main.fs:

```fsharp
module expectobenchmarktest
open Expecto

type ISerialiser =
  abstract member Serialise<'a> : 'a -> unit

type MySlowSerialiser() =
  interface ISerialiser with
    member __.Serialise _ =
      System.Threading.Thread.Sleep(30)

type FastSerialiser() =
  interface ISerialiser with
    member __.Serialise _ =
[...]

PR #384, opened 2020-04-17
Update BenchmarkDotNet Dependency

MNie:
- Update BenchmarkDotNet to 0.12.1 because of #381
- run paket update & install

Fixes: #381 
CC: @haf


haf:
Thank you @MNie !

PR #386, opened 2020-05-30
Support while loops inside test expressions et. al.

teo-tsirpanis:


haf:
Looking good. At this rate you'll find the ever-elusive progress-bar-related thread hang, too! Ready to merge?

teo-tsirpanis:
Yes, it's ready to merge.

> the ever-elusive progress-bar-related thread hang

What is its issue ID?

PR #394, opened 2020-07-22
list focused tests

### LINKED ISSUE #392: enhancement: list focused tests
To be sure the test suite runs all tests I use the `--fail-on-focused-tests` flag

```
$ dotnet run -- --fail-on-focused-tests
[10:13:11 ERR] It was requested that no focused tests exist, but 1 focused tests were found. <Expecto>
```
The output message is unhelpful. Luckily I can grep for `ftest`, but it is not ideal
Better were to see which tests are focused to unfocus them again.
Is it possible to list them instead of pointing out there are some?

Maybe also something similar for pending tests?

vilinski:
Fixes #392

### DIFF
--- a/Expecto/Expecto.Impl.fs
+++ b/Expecto/Expecto.Impl.fs
@@ -1043,9 +1043,15 @@ module Impl =
     if focused.Length = 0 then true
     else
       if config.verbosity <> LogLevel.Fatal then
+        let focusedTestNames =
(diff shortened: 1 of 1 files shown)


haf:
Thank you! 🔆

I would appreciate if you submit a PR towards the RELEASE_NOTES.md as well, with a description of your contribution.

PR #395, opened 2020-07-22
add options to list focused tests

vilinski:
following PR to enhancement #392  

Proposed changes and usage:
- `--list-tests` - lists all tests as before
- `--list-tests focused` - lists focused tests only
- `--list-tests pending focused` - lists pending and focused tests only
- added a "state char" before test name (N, P, F). It is hidden as before, if none of listed tests are focused or pending (not the case in Expecto.Tests)

```
P pending_test.case.one 
F focused_test.case.two
N normal_test.case.three 
```

### DIFF
--- a/Expecto/Expecto.Impl.fs
+++ b/Expecto/Expecto.Impl.fs
@@ -494,6 +494,8 @@ module Impl =
       /// An optional filter function. Useful if you only would
       /// like to run a subset of all the tests defined in your assembly.
       filter   : Test -> Test
+      /// List tests of specified state
+      listStates : FocusState list
       /// Allows the test printer to be parametised to your liking.
       printer : TestPrinters
       /// Verbosity level (default: Info).
@@ -528,6 +530,7 @@ module Impl =
         stressTimeout = TimeSpan.FromMinutes 5.0
         stressMemoryLimit = 100.0
         filter = id
+        listStates = []
         failOnFocusedTests = false
         printer =
           let tc = Environment.GetEnvironmentVariable "TEAMCITY_PROJECT_NAME"
--- a/Expecto/Expecto.fs
+++ b/Expecto/Expecto.fs
@@ -505,16 +505,18 @@ module Tests =
     tests
     |> Seq.iter (fun (stateChar, name) ->
       if hideState then
-        Printf.bprintf result "%s\n" name
-        //logger.info (Message.eventX "{name}" >> Message.setField "name" name)
+        Printf.bprintf result "\n%s" name
       else
-        // logger.info (Message.eventX "{stateChar} {name}"
-        //   >> Message.setField "stateChar" stateChar
-        //   >> Message.setField "name" name
-        // )
-        Printf.bprintf result "%s %s\n" stateChar name
+        Printf.bprintf result "\n%s %s" stateChar name
     )
-    logger.info (Message.eventX "{result}" >> Message.setField "result" (string result))
+    Printf.bprintf result "\n"
+
+    logger.logWithAck Info (
+      Message.eventX "{result}"
+      >> Message.setField "result" (string result)
+    )
+    |> Async.RunSynchronously
+    Printf.printfn "" // TODO workaround to flush output
 
   /// Prints out names of all tests for given test suite.
   let duplicatedNames (join: JoinWith) test =
--- a/Expecto/Expecto.fs
+++ b/Expecto/Expecto.fs
@@ -501,12 +501,20 @@ module Tests =
 
     let hideState = tests |> List.exists(fun (stateChar,_) -> stateChar <> "N") |> not
 
+    let result = System.Text.StringBuilder()
     tests
     |> Seq.iter (fun (stateChar, name) ->
       if hideState then
-        printfn "%s" name
+        Printf.bprintf result "%s\n" name
+        //logger.info (Message.eventX "{name}" >> Message.setField "name" name)
       else
-        printfn "%s %s" stateChar name)
+        // logger.info (Message.eventX "{stateChar} {name}"
+        //   >> Message.setField "stateChar" stateChar
+        //   >> Message.setField "name" name
+        // )
+        Printf.bprintf result "%s %s\n" stateChar name
+    )
+    logger.info (Message.eventX "{result}" >> Message.setField "result" (string result))
 
   /// Prints out names of all tests for given test suite.
   let duplicatedNames (join: JoinWith) test =
--- a/Expecto/Expecto.fs
+++ b/Expecto/Expecto.fs
@@ -486,10 +486,10 @@ module Tests =
 
   /// Prints out names of all tests for given test suite.
   let listTests config test =
-      Test.toTestCodeList test
-      |> Seq.filter(fun t -> List.isEmpty config.listStates ||
-                             List.contains t.state config.listStates)
-      |> Seq.iter (fun t -> printfn "%s" (config.joinWith.format t.name))
+    Test.toTestCodeList test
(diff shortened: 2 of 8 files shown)


Expecto/Expecto.fs
```diff
@@ -402,6 +408,8 @@ module Tests =
   type FillFromArgsResult =
     | ArgsRun of ExpectoConfig
     | ArgsList of ExpectoConfig
```
haf, line 410: Wouldn't these be better placed in the config?

Expecto/Expecto.fs
```diff
@@ -472,6 +482,10 @@ module Tests =
             Seq.fold (fun s a -> foldCLIArgumentToConfig a s) baseConfig cliArguments
           if List.contains List_Tests cliArguments then
             ArgsList config
```
haf, line 484: Inside this DU, that is?

vilinski:
Hope I understand you right. 
- added a field into ExpectoConfig
- changed the option `--list-tests` to optionally receive list of test states, added tests
- added entries into release notes for this and previous PR

also please check wording, my english is not that strong

Expecto/Expecto.Impl.fs
```diff
@@ -494,6 +494,8 @@ module Impl =
       /// An optional filter function. Useful if you only would
       /// like to run a subset of all the tests defined in your assembly.
       filter   : Test -> Test
+      /// List tests of specified state
```
haf, line 497: What does this mean? Is this a configuration flag?

haf: So this is a configuration flag to show only all tests that have one or two or three of the specified states?

Expecto/Expecto.fs
```diff
@@ -387,9 +384,7 @@ module Tests =
       "--filter-test-list", "Filters the list of test lists by a given substring.", Args.string Filter_Test_List
       "--filter-test-case", "Filters the list of test cases by a given substring.", Args.string Filter_Test_Case
       "--run", "Runs only provided list of tests.", Args.list Args.string Run
-      "--list-tests", "Don't run tests, but prints out list of tests instead.", Args.none List_Tests
-      "--list-focused-tests", "Don't run tests, but prints out list of focused tests instead.", Args.none List_Focused_Tests
-      "--list-pending-tests", "Don't run tests, but prints out list of pending tests instead.", Args.none List_Pending_Tests
+      "--list-tests", "Don't run tests, but prints out list of tests instead. Lists only tests with specified state(s), or all tests if not.", Args.list (Args.focusState) List_Tests
```
haf, line 387: Perhaps we should make it a `--summary-state` or something similar to the existing "display only" mnemonics?

RELEASE_NOTES.md
```diff
@@ -1,3 +1,7 @@
+#### 9.0.3 - 2020-07-27
```
haf, line 1: 👍

Expecto/Expecto.Impl.fs
vilinski: the proposed usage would be:

- `--list-tests` - lists all tests as before
- `--list-tests focused` - lists focused tests only
- `--list-tests pending focused` - lists pending and focused tests

but I'm of course open for any alternatives.

Expecto/Expecto.fs
vilinski: you mean let `--list-tests` alone but add `--summary-state` to display states?

vilinski: what you think about just add states (or/and maybe colors) to `--list-tests`?

```
P pending_test.case.one 
F focused_test.case.two
N normal_test.case.three 
```

haf: I didn't realise we had this flag from before; let's leave it alone; if it can work without an argument (as a default showing everything) that's good. Also good would be your latest suggestion about displaying state next to each test.

vilinski: Sorry for delay, I have added a "state char" before test names. Updated PR description. 👎 or 👍 ?

haf:
Build needs fixing and there's a `printf`: wouldn't it be better to go through the logger like everywhere else? How available is the logger where you've made your change?

vilinski:
> Build needs fixing

Oh, I see, there is a bug in NuGet/paket which I'm also struggling with. https://github.com/NuGet/Home/issues/9786
Workaround with commenting out a line in `paket.Restore.targets` didn't work. No clue how to fix this until it's done in NuGet

> and there's a `printf`: wouldn't it be better to go through the logger like everywhere else? How available is the logger where you've made your change?

`printfn` is a shadowed function which was used here before. `Impl.logger` is also awailable from `listTests` function. But I'm failed to figure out how to use it right. Is `logger.info` a right thing to do here? My last commit results in no output at all. 
actually it does if I append printfn after log or `Async.RunSynchronously`. Seems it's not get flushed before program ends.

haf:
@vilinski You can do an Async.RunSynchronously for now; the most important bit is that it goes through the logger so we can quelch output, should we wish to. I'm not too happy on shadowed stdlib functions.

vilinski:
`Async.RunSynchronously` and `logWithAck` doesn't help. it only works with `printfn ""` at end ...
An alternative would be to return the string from `listTests` instead of logging it.
also all other occurences of shadowed printfn are in the calling function `runTestsWithArgsAndCancel`.

haf:
The solution here is not to add a "fake" printf, but to hunt down why the logger's flush doesn't work. I'm afraid I don't have time to spare right now, but I would very much appreciate if you feel up for it.

vilinski:
tried to look into code and debug. `Monitor.Enter` and `Monitor.Exit` calls in progress printer are looking plausible for me.

however noticed that the `Logger.log` and `Logger.logWithAck` are identical in all leaf implementations `LiterateConsoleTarget`, `TextWriterTarget`, `OutputWindowTarget`

I have added a flush into `LiterateConsoleTarget`:

```
$ dotnet run --framework netcoreapp3.1 --  --list-tests pending
[00:37:38 INF] 
P findFastest.find fastest sleep
P all focused tests/ignored by attribute
P performance.same function is faster should fail
P findFastest.find fastest lo
P findFastest.find fastest hi
 <Expecto>

```

Comparet to no output at all for the above command it's a progress. 
But can you please confirm this is the right way to do? If not I'm afraid I can't understand how to fix this either.

Expecto/Expecto.fs
```diff
@@ -484,9 +485,37 @@ module Tests =
...
+    |> Seq.iter (fun (stateChar, name) ->
+      if hideState then
+        Printf.bprintf result "\n%s" name
+      else
+        Printf.bprintf result "\n%s %s" stateChar name
+    )
+    Printf.bprintf result "\n"
+
+    logger.logWithAck Info (
+      Message.eventX "{result}"
```
haf, line 515: Better to make the event the result IMO;

haf:
I think we can live with that flush call; if all other tests are passing. What happens if you disable progress printing? Flush still works, but is a no-op?

vilinski:
> I think we can live with that flush call; if all other tests are passing. 

No failing tests in `Expecto.Tests`.  What about other two `logWithAck`s? Do they also need a flush?

> What happens if you disable progress printing? 

`--no-spinner` option? I see the same output without progress spinner in either cases, maybe because there is just one log call for `--list-tests`

Still don't know hot to fix the build though

Expecto/Expecto.fs
vilinski: please explain, `Message.eventX result` without fields?

haf:
Thanks @vilinski ! 🎊

PR #399, opened 2020-11-05
Improved error msg of floats in floatClose

roboz0r:
When comparing small values the default `%f` formatter will not show a difference between actual and expected even when the test fails. Changed to `%.20g` to print the full precision of the value and automatically select the most compact representation.

### DIFF
--- a/Expecto/Expect.fs
+++ b/Expecto/Expect.fs
@@ -225,7 +225,7 @@ let floatClose accuracy actual expected message =
     failtestf "%s. Expected expected to not be infinity, but it was." message
   elif Accuracy.areClose accuracy actual expected |> not then
     failtestf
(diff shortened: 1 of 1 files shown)


haf:
Nice! 

I'm going to remove the Windows build, unless someone would like to maintain it? "PacketPack" task failed with no output is utterly befuddling.

baronfel:
This paket error is because the nuget team released a regression in their msbuild targets in the 3.1.4xx line of SDKs. You can fix it by staying on 3.1.3xx or moving to 5.0.1xx-rc2 or later SDKs.

haf:
Of course they did. Has that team ever brought anything good into this world?

haf:
But I don't think it's that easy because AppVeyor force-upgraded the SDK to 4xx.

haf:
Thanks @roboz0r ! 🎊

PR #402, opened 2020-11-10
Changes required to support a .net5.0 build of expecto.

baronfel:
This does the package updates and TFM additions required to fully support .net 5 in Expecto.  It's really straightforward, the only potential problems might be:

* CI support (github actions works if that's acceptable?)
* prerelease of FAKE packages and runtime was required to get the a compatible version of the Fake/Paket integration to understand .net 5)


Expecto.Hopac/Expecto.Hopac.fsproj
```diff
@@ -11,6 +11,9 @@
   <ItemGroup>
     <ProjectReference Include="..\Expecto\Expecto.fsproj" />
   </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'net461'">
```
baronfel, line 14: this change was to address a build error when building this project against net461. It seems that System.Threading.Tasks wasn't included by default somewhere.

Expecto/Expecto.Impl.fs
```diff
@@ -289,7 +289,7 @@ module Impl =
           let spirit =
             if summary.successful then "Success!" else String.Empty
           let commonAncestor =
-            let rec loop ancestor (descendants : string list) =
+            let rec loop (ancestor: string) (descendants : string list) =
```
baronfel, line 292: this change was to address an inference error because `ancestor` has a member indexed in an unknown way due to new overloads in .net 5.

haf:
Thanks @baronfel !

PR #408, opened 2021-01-10
prevent globbing

despairblue:
I saw that you mentioned this in your video explicitly. Thought it makes sense to have the README consistent with that. Also this tripped me when I copy and pasted this :)


haf:
Thanks @despairblue !

Issue #419, opened 2021-04-27
Expecto.Diff seems to mix up actual and expected


lydell:
Repro repo: https://github.com/lydell/ExpectoDiffMixup

I’ve enabled `Expecto.Diff`:

```fs
Expect.defaultDiffPrinter <- Diff.colourisedDiff
```

Running this assertion:

```fs
Expect.equal "actual text" "expected text" ""
```

… results in this message:

```
. String does not match at position 0. Expected char: 'e', but got 'a'.
---------- Actual: --------------------
expected text
---------- Expected: ------------------
actual text
```

The `Expected char: 'e', but got 'a'.` is correct, but the part after is mixed up. I would expect this:

```
. String does not match at position 0. Expected char: 'e', but got 'a'.
---------- Actual: --------------------
actual text
---------- Expected: ------------------
expected text
```

Screen shot:

<img width="572" alt="image" src="https://user-images.githubusercontent.com/2142817/116296671-c6f9e480-a79a-11eb-8ec6-9f188c04fe3b.png">

Also note how the blue color leaks into the  `-- Actual: --` line. It’s even more apparent if both `actual` and `expected` start with some common text – notice how the first “some” is blue, while it should be white:

<img width="585" alt="image" src="https://user-images.githubusercontent.com/2142817/116296877-01638180-a79b-11eb-83b9-672335aca3cc.png">

It’s worse if you use `Expecto.Diff.equals` in `Xunit` using `dotnet test`:

<img width="821" alt="image" src="https://user-images.githubusercontent.com/2142817/116297073-34a61080-a79b-11eb-91d2-cff52c0f3347.png">

Notice how the red color leaks into the diff and colors the first “some” red.

Pinging @drhumlen in case you’re interested 🙂

PR #420, opened 2021-04-28
Fix flipped/mixed up Expecto.Diff formatting

lydell:
Fixes https://github.com/haf/expecto/issues/419

I tested this by running `dotnet run --project Expecto.Sample --framework net5.0` in this repo.

First, the default output for comparison:

image

Notes:

- Expected comes first, then actual.
- Expected is in green, actual in red.

Then, the `Expecto.Diff` output:

image

Notes:

- The Actual/Expected _headings_ are flipped compared to the default formatting.
- The actual/expected _text_ are still in the same order as the default formatting. In other words, the headings are for the wrong text.
- The colors are flipped compared to the default formatting.
- Also note how the first heading and `{A =` is blue, while it should be white.

Finally, the output with this PR:

image

Notes:

- Fixed the headings so Expected comes first, just like in the default formatting. And now the headings and the actual/expected text match.
- Fixed the colors. Expected is now green and actual is red, just like the default formatting.
- Fixed the heading and start-of-text color. It’s now reset.

### DIFF
--- a/Expecto.Diff/Library.fs
+++ b/Expecto.Diff/Library.fs
@@ -21,7 +21,7 @@ let colourisedDiff actual expected =
     | ChangeType.Deleted -> ColourText.colouriseText ConsoleColor.Red text
     | ChangeType.Modified -> ColourText.colouriseText ConsoleColor.Blue text
     | ChangeType.Imaginary -> ColourText.colouriseText ConsoleColor.Yellow text
-    | ChangeType.Unchanged | ChangeType.Imaginary | _ -> text
+    | ChangeType.Unchanged | _ -> text
 
   let colouriseLine (line: DiffPiece) =
     if line.SubPieces.Count = 0 then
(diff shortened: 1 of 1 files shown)


lydell:
@haf Friendly ping :) Small change, great quality of life improvement

haf:
Ah sorry, haven't seen this. Thanks @lydell !

lydell:
Thanks! What’s the plan for the next release? Is it days, weeks, months? Easier to wait for something good when you’ve got a rough estimate of the time frame 🙏

olivercoad:
@haf Any chance you could push an update to nuget some time?

haf:
@olivercoad Just a lot to do every day. Could you help me get this auto-publishing with Github Actions? I'd happily add in secrets and such to the repo so that it can work.

haf:
I've added the secret NUGET_RELEASE_KEY that we can use to push new nugets on every commit to master. (And let's rename it to `main` while we're at it!)

olivercoad:
No worries. Thanks for setting up the secret, I'll try and have a look at how to get auto-publishing working.

olivercoad:
> push new nugets on every commit to master

I'm guessing something like this would do that (looking at https://docs.github.com/en/actions/guides/building-and-testing-net#publishing-to-package-registries)

```yaml
name: Publish Nuget Packages
on:
  push:
    branches: [ main ]

jobs:
  build:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v2
    - name: Setup .Net
      uses: actions/setup-dotnet@v1
      with:
        dotnet-version: "5.0"
      env:
        NUGET_TOKEN: ${{secrets.NUGET_RELEASE_KEY}}
    - run: ./fake build --target Push
```

Though you'd probably want some kind of manual action so it it only pushes when you change the version number or something.

baronfel:
you can also do manual job runs, which we do in fsautocomplete: 

```
on:
  workflow_dispatch:
    inputs:
      username:
        description: Github username of the person triggering the release
        default: "baronfel"
        required: true
      email:
        description: Email of the person triggering the release
        default: "some_email@place.com"
        required: true
```

these can be referenced in subsequent steps, or you can leave them out entirely to require a manual job.  in either case only users with write access can trigger them.

haf:
I'm happy with any solution that gives me more time.

Every time I run `./fake build --target Push` that target fails for unspecified reasons and I have to manually push everything with the shell script line that is a comment in the fsx file.

haf:
I changed the script like you suggested @olivercoad

PR #423, opened 2021-07-30
Feature - Run tests from many assemblies

benguenter:
When there are multiple assemblies of tests for an FSharp project, it is a pain to run each project of tests individually using the `dotnet` cli.
I think it's valuable to use the `dotnet watch` tooling as well which doesn't work as nicely with multiple test projects either.
This PR adds the ability to specify multiple assemblies from the test runner in the `Main.fs` file when calling `runTestsInAssembly`.
That way all the results show up in one big list and we can use useful tooling like `dotnet watch` on a single project.

### DIFF
--- a/Expecto/Expecto.Impl.fs
+++ b/Expecto/Expecto.Impl.fs
@@ -1026,16 +1026,31 @@ module Impl =
       SourceLocation.empty
 
   /// Scan filtered tests marked with TestsAttribute from an assembly
-  let testFromAssemblyWithFilter typeFilter (a: Assembly) =
+  let testListFromAssemblyWithFilter typeFilter (a : Assembly) =
     a.GetExportedTypes()
     |> Seq.filter typeFilter
     |> Seq.choose testFromType
     |> Seq.toList
+
+  /// Scan filtered tests marked with TestsAttribute from an assembly
+  let testFromAssemblyWithFilter typeFilter (assembly : Assembly) =
+    testListFromAssemblyWithFilter typeFilter assembly
+    |> listToTestListOption
+
+  /// Scan filtered tests marked with TestsAttribute from multiple assemblies
+  let testFromAssembliesWithFilter typeFilter (assemblies : Assembly seq) =
+    assemblies
+    |> Seq.map (testListFromAssemblyWithFilter typeFilter)
+    |> List.concat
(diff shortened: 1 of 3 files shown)


Expecto/Expecto.Impl.fs
```diff
@@ -1026,17 +1026,32 @@ module Impl =
...
+  let testListFromAssemblyWithFilter typeFilter (a : Assembly) =
     a.GetExportedTypes()
     |> Seq.filter typeFilter
     |> Seq.choose testFromType
     |> Seq.toList
+
+  /// Scan filtered tests marked with TestsAttribute from an assembly
+  let testFromAssemblyWithFilter typeFilter (assembly : Assembly) =
+    testListFromAssemblyWithFilter typeFilter assembly
+    |> listToTestListOption
```
haf, line 1038: This looks like it just changed the return type of the function `testFromAssemblyWithFilter` from a list to whatever `listToTestListOption` returns?

benguenter: True, we could probably compose those functions together instead. 👍

haf:
Is this ready to merge?

benguenter:
> Is this ready to merge?

Sure is!

Expecto/Expecto.Impl.fs
haf: My point here is that you've changed a public ABI, by currying instead of providing explicit parameters. Could you explicitly declare and pass the parameters please to avoid this change? (Eta expansion) Thank you

Expecto/Expecto.fs
```diff
@@ -618,6 +618,15 @@ module Tests =
     let tests = testFromThisAssembly() |> Option.orDefault (TestList ([], Normal))
     runTestsWithArgsAndCancel ct config args tests
 
+  /// Runs tests in the specified assemblies with the supplied command-line options.
+  /// Returns 0 if all tests passed, otherwise 1
+  let runTestsInAssembliesWithCLIArgsAndCancel (ct:CancellationToken) cliArgs args assemblies =
```
ratsclub, line 623: This is mostly a style-related comment. :stuck_out_tongue_winking_eye: 

The other functions have the signature written as `let foo (bar : baz) = ...`. Should this one follow the same style?

PR #427, opened 2021-12-03
Update F# Version to most current (6)

### LINKED ISSUE #424: F# 5 String Interpolation Causes Spurrious Test Failures
If a method under test uses F# 5 string interpolation (e.g. `sprintf $"{myVar}"`), the test will throw a method not found error and fail, whether the actual outcome of the test was a failure. This is due to Expecto pinning F# Core 4 in its `paket.dependencies`. 

I've forked the project, updated the F# Core version to 5.0.2, and run both the build and the test suite; both pass with no errors. I'm very happy to PR that change, but I wanted to ask first: are there any other checks, build steps, or other things I should verify to make sure the change is safe? Very happy to help however I can. 

Thanks so much for Expecto! It really is a delight to use.

Gastove:
I got interested in this because of #424 -- without F# 5, Expecto can't test any
assemblies that use F# 5 string interpolation. Since I opened that issue, F# 6
has come out. Might as well use that!

My only concern here is testing. Running the test suite, I see two reported failures, both having to do with the performance of SHA256 vs MD5. However, both of those tests also fail locally on main for me, so I don't think it's related to this change. 

Thanks for the lovely testing library!

Closes #424

### DIFF
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -3,7 +3,7 @@ framework net461, netstandard2.0, netcoreapp3.1, net5.0
 storage none
 version 5.245.1
 
-nuget FSharp.Core 4.6.0
+nuget FSharp.Core ~6
 nuget FsCheck ~> 2.14
 nuget Hopac ~> 0.4
 nuget DiffPlex ~> 1.5


64J0:
Any update on this?

PR #439, opened 2022-09-18
update to current naming scheme

ShalokShalom:
I changed dotnet core to dotnet and updated the version number in one example from 5.0 to 6.0


haf:
Thanks 🙏

Issue #429, opened 2022-01-28
Mark deprecated functions with ObsoleteAttribute


charlieturndorf:
I've been using `runTestsInAssembly` for a long time, and just now found out it's deprecated when I went to bootstrap a new test project. No complaints about the deprecation! I think this is a positive change.

Though if functions are marked `[<Obsolete("Deprecated, please use... instead")>]` then people will see a warning and have a chance to update their code before a breaking change hits.

Anyway team, thanks for your hard work on this excellent testing framework!

PR #446, opened 2023-03-22
Add obsolete attribute on deprecated functions

ratsclub:
This PR introduces Obsolete attributes on deprecated functions. Ref: #429 

Also, we could perhaps remove them altogether on a future major release. What do you think?


haf:
Yes, let's remove them in the next-next major

PR #447, opened 2023-03-27
Bump minimum target framework

ratsclub:
This PR bumps the minimum target framework to the current LTS. This makes it possible to use Tasks.


Expecto.Tests/Tests.fs
```diff
@@ -124,8 +123,8 @@ let tests =
             let greenStart = ANSIOutputWriter.colourText (ANSIOutputWriter.getColour()) ConsoleColor.Green
             let greenEnd = ANSIOutputWriter.colourText (ANSIOutputWriter.getColour()) ConsoleColor.Cyan
             "Failing - record with different content.\nRecord does not match at position 2 for field named `b`. Expected field with value: \"dw\", but got \"de\".\n"+
-            greenStart+"expected"+greenEnd+":\n{a = \"dd\";\n b = \"d"+greenStart+"w"+greenEnd+"\";}\n"+
-            redStart+"  actual"+redEnd+":\n{a = \"dd\";\n b = \"d"+redStart+"e"+redEnd+"\";}"
+            greenStart+"expected"+greenEnd+":\n{ a = \"dd\"\n  b = \"d"+greenStart+"w"+greenEnd+"\" }\n"+
+            redStart+"  actual"+redEnd+":\n{ a = \"dd\"\n  b = \"d"+redStart+"e"+redEnd+"\" }"
```
ratsclub, line 127: This is something that I could not track why I had to change. The difference is the following:

The previous version printed in this format here:
```fs
{a = "dd";
 b = "dw";}
  actual:
{a = "dd";
 b = "de";}
```

However, the new one expects this format:
```fs
{ a = "dd"
  b = "de" }
```

ratsclub: As far as I can tell, yes. It does use `%A`... Is there a better approach here?

Expecto.BenchmarkDotNet.Tests/Expecto.BenchmarkDotNet.Tests.fsproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>net5.0;netcoreapp3.1;net461</TargetFrameworks>
+    <TargetFrameworks>netstandard2.1;net6.0</TargetFrameworks>
```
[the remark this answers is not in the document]
ratsclub: I'm kind of new to how dotnet versioning works. I put `netstandard2.1` due to [this table](https://learn.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-1#select-net-standard-version) here. Do you think that it's redundant to keep it and instead I should go with plain `net6.0` here?

Expecto.Tests/Tests.fs
baronfel: that's a fine approach! it just also explains why the format differed. you're essentially off-loading the responsibility for formatting, so you just might have to change your test expectations every so often :shrug:

ratsclub:
As far as I can tell, yes. It does use `%A`... Is there a better approach here?

Expecto.BenchmarkDotNet.Tests/Expecto.BenchmarkDotNet.Tests.fsproj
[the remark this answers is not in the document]
baronfel: NS2.1 is a perfectly fine choice, but it does lock you out of .NET Framework support entirely. If you're ok making that decision, then you may as well limit to only in-support .NET versions, which you can get with a single TFM (at this moment) targeting net6.0.

ratsclub:
@baronfel @haf this is building, do you think this is good enough? My idea is to tackle as much issues as possible (such as #446 and #383) and release a new major with the breaking changes.

1eyewonder:
Would this also be a fix to this issue as well? #440

ratsclub:
@1eyewonder This PR bumps the dependency, so I would say yes.

ratsclub:
@haf This is ready for review. Do you think that I should change anything else here?

PR #450, opened 2023-04-18
Migrate to FsCheck3 in a separate project

farlee2121:
## Motivation

I'd like to FsCheck version 3 for some of its improvements, but Expecto currently doesn't work with FsCheck 3.

### Separate Migrated Assembly
This PR adds a new project with FsCheck 3 support that we could publish separately. FsCheck 3 may not reach full-release any time soon. This way users can opt-in to FsCheck3 without disrupting the mainline Expecto.FsCheck project.

An alternative solution would be to create pre-release versions of the Expecto.FsCheck package using a `-fscheck3` suffix to differentiate from the normal pre-release packages. This option may require maintaining a special branch. It might also be possible to leave the separate projects but use the same package id and publish special pre-release versions from the migrated project. However, it's unclear how version selection may work for users that specify non-deterministic pre-release versions.
In short, the pro is that we don't have to manage a separate package and eventually deprecate it. The con is a potentially confusing version experience.

### Migration changes

The migration itself was pretty simple.
- FsCheck's configuration flow changed to a builder format. 
- Replay was changed to a new type, but we can map our current format into the new one.
- The names and a few parameters of test result union cases were changed, but updating the mapping was simple

No new FsCheck features are directly leveraged in the current migration.

### Breaking Changes

FsCheck changed the way they handle random values, so the same Repeat seed does not produce consistent results across verions. Thus Expecto's unit tests depending on a deterministic result cannot pass on FsCheck 2 & 3 without modification.

### DIFF
--- a/Expecto.FsCheck3/Expecto.FsCheck3.fsproj
+++ b/Expecto.FsCheck3/Expecto.FsCheck3.fsproj
@@ -1,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <PackageId>Expecto.FsCheck3</PackageId>
+    <Description>Property testing for Expecto, powered by FsCheck3</Description>
+    <TargetFramework>net6.0</TargetFramework>
+  </PropertyGroup>
+  <ItemGroup>
+    <Compile Include="FsCheck3.fs" />
+    <None Include="paket.references" />
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\Expecto\Expecto.fsproj" />
+  </ItemGroup>
+  <Import Project="..\NuGet.props" />
+  <Import Project="..\.paket\Paket.Restore.targets" />
+</Project>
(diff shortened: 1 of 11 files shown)


farlee2121:
Hmm. I maybe squashed the changes too aggressively. It's hard to pick out the migration changes from the copying of the projects and tests.
Should I try to split those commits out?

haf (approved):

farlee2121:
@ratsclub [You mentioned](https://github.com/capivarasdev/patronum/pull/1#issuecomment-1582698221) fixes to the build system, does that include packing? 
I noticed there is a Fake target for pack, but it doesn't seem to be working for me.

ratsclub:
I intend to write the new Fake scripts over the weekend. In the meantime, let me see if I can make it work for you.

Also, @haf can you provide one of us access to this repo's secret? This way we could setup a CI job to do releases.

ratsclub:
@farlee2121 It's now published, thanks!

farlee2121:
@ratsclub Thanks for updating the build scripts!

PR #455, opened 2023-05-27
Add missing Flip.Expect.hasLength

JohnTheGr8:
also adjust the documentation of `Flip.Expect.stringContains` and `Flip.Expect.stringHasLength`


farlee2121:
@JohnTheGr8 Your changes look good, but it looks like there's some mischief going on with the merge. Could you rebase from main?

PR #457, opened 2023-06-08
Update testTheoryWithResult example with testTheory equivalent

farlee2121:
testTheoryWithResult was removed because the same can be accomplished with testTheory. It cuts down the API surface considerably when considering focus, ignore, and async varients


ratsclub (approved):
Thanks, I forgot to remove those!

PR #459, opened 2023-06-20
Loosen FSharp.Core version requirement

### LINKED ISSUE #458: Loosen FSharp.Core version requirement
Expecto 10.0.0 requires FSharp.Core = 7.0.200 (exact version). It should be `>=` (minimum version). I am using 7.0.300 in all projects, and therefore can't update Expecto.

JohnTheGr8:
quick PR to change expecto's `FSharp.Core` version requirement from `(= 7.0.200)` to ` (>= 7.0.200 && < 8.0.0)`

closes #458

### DIFF
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -3,7 +3,7 @@ framework netstandard2.1, net6.0
 storage none
 version 7.1.5
 
-nuget FSharp.Core ~6
+nuget FSharp.Core ~> 7
 nuget FsCheck ~> 2.14
 nuget Hopac ~> 0.4
 nuget DiffPlex ~> 1.5
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -3,7 +3,7 @@ framework netstandard2.1, net6.0
 storage none
 version 7.1.5
 
-nuget FSharp.Core ~> 7
+nuget FSharp.Core >= 6
 nuget FsCheck ~> 2.14
 nuget Hopac ~> 0.4
 nuget DiffPlex ~> 1.5


JohnTheGr8:
Please let me know if I should take care of the failing security check as part of this PR...

farlee2121:
I did a bit of investigating, and the insecure package isn't actually a result of this change. It stems from some build-only dependencies, but the insecure transitive dependency can't currently be resolved just by upgrading direct package dependencies.

I think we're good to solve the two issues separately. However, looking at previous version constraints, I think `nuget FSharp.Core >= 6` is probably the constraint we're looking for. I'll let you grab that if you still want the merge. I appreciate you taking initiative to contribute!

JohnTheGr8:
> It stems from some build-only dependencies, but the insecure transitive dependency can't currently be resolved just by upgrading direct package dependencies.

Yup, I figured as much... You can trick Paket to update just that package by adding it to `paket.dependencies`, updating it and then resetting `paket.dependencies`.

> I think `nuget FSharp.Core >= 6` is probably the constraint we're looking for.

I tried this initially but it makes no difference: when packing, the minimum required version will be set to what Expecto uses, currently version `7.0.200`

farlee2121:
I thought about editing the lock file or similar, but it seems like an unstable solution.

Hmm. When I inspected the nuget packages, the `~>` approach does result in a max allowed version, while the `>=` approach does not
image
image

This seems more in line with [previous versioning constraints](https://www.nuget.org/packages/Expecto/9.0.4#dependencies-body-tab)

JohnTheGr8:
image

This is what I'm getting...

farlee2121:
Weird. Not sure why we'd get different results.

JohnTheGr8:
are you running an outdated version of nuget package explorer by any chance? `.NETCoreApp, Version=v6.0` sounds very odd 😂

JohnTheGr8:
anyways, just meant to point out that in my testing setting it to `>= 6` and `>= 7` made no difference as it would always result in `>= 7.0.200`... doesn't really matter

farlee2121:
Noted

PR #465, opened 2023-06-22
Fix paket install (FsCheck3 versioning)

farlee2121:
Replace #464. I updated recommended packaged, but whitebolt kept saying the new commit didn't warrant a rescan.

### DIFF
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -13,7 +13,7 @@ nuget BenchmarkDotNet ~> 0.13.5
 group FsCheck3
   source https://api.nuget.org/v3/index.json
 
-  nuget FsCheck ~> 3
+  nuget FsCheck >= 3 prerelease
 
 group Build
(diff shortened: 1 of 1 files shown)


farlee2121:
@JohnTheGr8 At last WhiteSource is happy and the paket issues should be resolved now

JohnTheGr8:
Nice one, would be nice if we could get a release out now... 

Note that, currently, running `paket outdated` on existing projects gives weird results because of #458. More specifically:
* projects on expecto 9.0.4 & FSharp.Core 7.0.300 will not see the update to expecto 10.0
* projects on expecto 10.0 & FSharp.Core 7.0.200 are prompted to downgrade expecto and upgrade FSharp.Core (which is interesting)

farlee2121:
Indeed. Had to figure out some bit rot with the release system first though. 
10.1.0 is live on nuget now

PR #466, opened 2023-06-22
Fix build dependency vulnerabilities

farlee2121:
CVE-2023-29331 and CVE-2023-29337

PR #467, opened 2023-06-25
Fix build dependency vulnerabilities 

farlee2121:
Fix CVE-2023-29331 and CVE-2023-29337

Replace #466 since vulnerability scan is not running for unclear reasons. Hopefully this will get it to run.


farlee2121:
closes #466
closes #463

PR #471, opened 2023-08-04
Added runTests to simplify usage in interactive mode

sporring:
In interactive mode, it is superfluous to give default empty parameters. Thus I propose runTests

### DIFF
--- a/Expecto/TestExpecto.fsx
+++ b/Expecto/TestExpecto.fsx
@@ -1,0 +1,8 @@
+#r "nuget:Expecto"
+open Expecto
+
+let simpleTest =
+  testCase "A simple test" <| fun () ->
+    Expect.equal 4 (2+2) "2+2 = 4"
+
+runTestsWithCLIArgs [] [||] simpleTest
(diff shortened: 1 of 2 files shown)


farlee2121:
This is something that caused me to double take when I first started with the library.

I feel like this is a fairly high-traffic API decision though, and @haf should have the final call.

Expecto/TestExpecto.fsx
```diff
@@ -0,0 +1,8 @@
+#r "nuget:Expecto"
+open Expecto
+
+let simpleTest =
+  testCase "A simple test" <| fun () ->
+    Expect.equal 4 (2+2) "2+2 = 4"
+
+runTestsWithCLIArgs [] [||] simpleTest
```
farlee2121, line 8: @sporring Could you clarify the motivation of these fsx files?

Expecto/Expecto.fs
```diff
@@ -633,6 +633,10 @@ module Tests =
   let runTestsWithCLIArgs cliArgs args tests =
     runTestsWithCLIArgsAndCancel CancellationToken.None cliArgs args tests
 
+  /// Runs all given tests.
+  /// Returns 0 if all tests passed, otherwise 1
+  let runTests = runTestsWithCLIArgs [] [||]
```
farlee2121, line 638: We should also document this in the ReadMe

Issue #477, opened 2023-09-07
Add helpers for nullable value types to `Expect` module


bisen2:
Currently, there are helper functions in the `Expect` module named `isNull` and `isNotNull` for checking if a reference type is null. I think that it would be beneficial to have similar helper functions for  `Nullable<'t>`. It is currently possible to work around this either by using `Expect.isTrue x.HasValue` in your test or defining the helper functions for yourself, but it seems to me like this would be a good thing to have built in.

Currently, I am defining the helper functions in my test library like so:
```fs
module Expect =
  let isNullValue (x: Nullable<'t>) message =
    if not x.HasValue then ()
    else failtestf "%s. Expected null, but was %A." message x
  let isNotNullValue (x: Nullable<'t>) message =
    if x.HasValue then ()
    else failtestf "%s. Expected non-null, but was null." message
```

I would be happy to PR this if it is something you would be interested in. Thanks!

haf:
Sounds good :) I generally don't mind adding more assert functions to the Expect module.

PR #478, opened 2023-09-19
Add `Nullable<'t>` helpers to `Expect` module

bisen2:
This PR adds the helper functions suggested in #477 for asserting if a `Nullable<'t>` wraps a null value or not.


Expecto.Tests/Tests.fs
```diff
@@ -69,6 +69,16 @@ let tests =
       ) |> assertTestFails
     ]
 
+    testList "null value tests" [
```
farlee2121, line 72: Should we test the passing cases too?

farlee2121:
Sorry this slipped through the cracks. It's just been release with 10.2.0

Issue #481, opened 2023-11-03
ColourLevel can only be set once


farlee2121:
I'm trying to improve the experience in interactive environments, and I noticed that [`ANSIOutputWriter.setColourLevel`](https://github.com/haf/expecto/blob/ef7ccd67d61498cb781f6a6d0ee1d271f44d5ad0/Expecto/Logging.fs#L882C28-L882C28) only allows setting once. This means the colour setting is only respected on the first call to a runTests* method, unless the process is restarted.

It looks like @AnthonyLloyd worked on this. Do you remember why it works that way?

Issue #365, opened 2019-11-14
How to show test results in interactive


halcwb:
How can I print the test results in a nice way in the interactive? I tried something like:

```
let config =
    { defaultConfig with ``parallel`` = false }

let handler (sum : Impl.TestRunSummary) =
    for (t, r) in sum.results do
        printfn "Test: %A Result: %A" t.name r.result

ValueRange.tests
|> runTests (config.appendSummaryHandler handler)

```
But this results in garbled output in the interactive. Is there an easy way to accomplish this?

haf:
You have to create a logger:

```fsharp
open Expecto.Logging
open Expecto.Logging.Message
let logger = Log.create "FSI"
logger.info (eventX "Hi")
```

halcwb:
@haf Sorry, but still not get this. For example with the following I was expecting that a verbose print should be visible in the FSI:

```fsharp
runTestsWithCLIArgs [ CLIArguments.Debug; CLIArguments.Sequenced ] [||]
```

However, whether or not I add the CLIArguments the print out just shows:

> [15:43:56 INF] EXPECTO? Running tests... <Expecto>
> [15:43:56 INF] EXPECTO! 28 tests run in 00:00:00.0477646 for PIM – 28 passed, 0 ignored, 0 failed, 0 errored. Success! <Expecto>
> val it : int = 0
> 

I use VSCode on a Macbook.

farlee2121:
@halcwb 

It's still a bit scrappy, but I wrote these run methods for better interactive support. They capture the logs normally printed to the console and return them as a string from the run method. There's one with ANSI coloration and one without

```fsharp
module Tests =

    let private literateOutputWriter (outputBuilder: StringBuilder) (text: (string*ConsoleColor) list) : unit =
            let colorizeLine (text, color) = ColourText.colouriseText color text
            let sbAppend (builder: StringBuilder) (text: string) =
                builder.Append(text)

            text
            |> List.iter (colorizeLine >> (sbAppend outputBuilder) >> ignore)
            

    let private colorlessOutputWriter (outputBuilder: StringBuilder) (text: (string*ConsoleColor) list) : unit =
        text
        |> List.iter (fun (text, color) ->
            outputBuilder.Append(text)
            |> ignore)


    let run_ReturnLogs cliArgs args tests =

        let outputBuilder = StringBuilder("")
        
        Global.initialise 
            { Global.defaultConfig with
                getLogger = fun name ->
[...]

PR #482, opened 2023-11-06
Improve support for interactive experiences

farlee2121:
## Motivation
Expecto is uniquely suitable for interactive environments (like FSI or Polyglot Notebooks) because it treats tests as values. Tests can be defined and executed all in the same code.

However, none of the current runTests methods print output to interactive environments like FSI.

## If this PR is merged
This PR adds a new method that collects logs from the test run then returns them as a string.
This allows easily displaying test output in interactive environments.

For example, here's a polyglot notebook using `runTestsWithCLIArgs`

image

Here's the same interactive cell using `runTestsReturnLogs`
image

The new `runTestsReturnLogs` should obey options the same as other run methods
image

### DIFF
--- a/Expecto/Expecto.fs
+++ b/Expecto/Expecto.fs
@@ -644,6 +644,9 @@ module Tests =
   let runTestsInAssemblyWithCLIArgs cliArgs args =
     runTestsInAssemblyWithCLIArgsAndCancel CancellationToken.None cliArgs args
 
+  /// Runs all given tests with the supplied typed command-line options.
+  /// Returns the console output as a string (with ANSI coloring by default)
+  /// Useful for interactive environments like F# interactive or notebooks
   let runTestsReturnLogs cliArgs args tests =
     let tryBuildConfig cliArgs args =
         let config =
--- a/Expecto/Expecto.fs
+++ b/Expecto/Expecto.fs
@@ -643,3 +643,44 @@ module Tests =
   /// Returns 0 if all tests passed, otherwise 1
   let runTestsInAssemblyWithCLIArgs cliArgs args =
     runTestsInAssemblyWithCLIArgsAndCancel CancellationToken.None cliArgs args
+
+  let runTestsReturnLogs cliArgs args tests =
+    let tryBuildConfig cliArgs args =
+        let config =
+          Seq.fold (fun s a -> foldCLIArgumentToConfig a s)
+            ExpectoConfig.defaultConfig cliArgs
+
+        match ExpectoConfig.fillFromArgs config args with
+        | ArgsUsage (usage, errors) ->
+          None
+        | ArgsList config 
+        | ArgsRun config 
+        | ArgsVersion config ->
+          Some config
+
+    let literateOutputWriter (outputBuilder: Text.StringBuilder) (text: (string*ConsoleColor) list) : unit =
+      let colorizeLine (text, color) = ColourText.colouriseText color text
+      let sbAppend (builder: Text.StringBuilder) (text: string) =
+        builder.Append(text)
+
+      text
+      |> List.iter (colorizeLine >> (sbAppend outputBuilder) >> ignore)
+
+    let outputBuilder = System.Text.StringBuilder("")
+
+    let verbosity = 
+      tryBuildConfig cliArgs args 
+      |> Option.defaultValue ExpectoConfig.defaultConfig 
+      |> (fun config -> config.verbosity)
+        
+    Global.initialise 
+      { Global.defaultConfig with
+          getLogger = fun name ->
+            Expecto.Logging.LiterateConsoleTarget(
+              name, 
+              minLevel = verbosity, 
+              outputWriter = (literateOutputWriter outputBuilder)) :> Expecto.Logging.Logger
+      }
+
+    runTestsWithCLIArgs cliArgs args tests |> ignore
+    outputBuilder.ToString()
--- a/Expecto/Expecto.fs
+++ b/Expecto/Expecto.fs
@@ -360,6 +360,10 @@ module Tests =
   type SummaryHandler =
     | SummaryHandler of (TestRunSummary -> unit)
 
+  [<ReferenceEquality>]
+  type LoggingConfigFactory = 
+    | FromVerbosity of (LogLevel -> LoggingConfig)
+
   /// The CLI arguments are the parameters that are possible to send to Expecto
   /// and change the runner's behaviour.
   type CLIArguments =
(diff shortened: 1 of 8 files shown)


Expecto/Logging.fs
```diff
@@ -879,7 +879,7 @@ module internal ANSIOutputWriter =
     override __.WriteLine() = write "\n"
 
   let mutable internal colours = None
-  let internal setColourLevel c = if colours.IsNone then colours <- Some c
+  let internal setColourLevel c = colours <- Some c
```
farlee2121, line 882: I couldn't discern a reason this needed to be settable only once and it prevents the colour setting from working as expected in interactive environments

farlee2121: #481

Expecto/Expecto.fs
```diff
@@ -643,3 +643,44 @@ module Tests =
   /// Returns 0 if all tests passed, otherwise 1
   let runTestsInAssemblyWithCLIArgs cliArgs args =
     runTestsInAssemblyWithCLIArgsAndCancel CancellationToken.None cliArgs args
+
+  let runTestsReturnLogs cliArgs args tests =
```
farlee2121, line 647: I could use feedback on this method name.

Expecto/Expecto.fs
```diff
@@ -643,3 +643,44 @@ module Tests =
...
+    let tryBuildConfig cliArgs args =
+        let config =
+          Seq.fold (fun s a -> foldCLIArgumentToConfig a s)
+            ExpectoConfig.defaultConfig cliArgs
+
+        match ExpectoConfig.fillFromArgs config args with
```
farlee2121, line 653: I believe this results in double args parsing, but it seemed like the least complex option (and easiest to follow) and we're not doing that much parsing generally. It also only applies to this interactive run method

Expecto/Expecto.fs
farlee2121: Maybe `runTestsForInteractive`

ratsclub:
I will review this later today!

README.md
```diff
@@ -250,6 +251,22 @@ Signature `CancellationToken -> CLIArguments seq -> string[] -> int`. Runs the t
 assembly and also overrides the passed `CLIArguments` with the command line
 parameters. All tests need to be marked with the `[<Tests>]` attribute.
 
+### `runTestsReturnLogs`
+
+Signature `CLIArguments seq -> string[] -> Test -> string`.
+Accepts the same arguments as `runTestsWithCLIArgs`, but returns the console output as a string.
+
+This is useful for interactive environments like [F# interactive](https://learn.microsoft.com/en-us/dotnet/fsharp/tools/fsharp-interactive/) and [Polyglot Notebooks](https://marketplace.visualstudio.com/items?itemName=ms-dotnettools.dotnet-interactive-vscode), where console output is not available but the returned string can be displayed instead.
+
```
farlee2121, line 260: I plan to add a screen shot once I can show the proper package reference.
Something like `#r "nuget: Expecto, 10.2.0"`.

ratsclub:
One question, instead of expanding the API with a function that only changes the logging output, wouldn't it be better to add an option that tells where the output should go?

farlee2121:
In short, I don't think we can generally redirect output for interactive environments. So, configuration creates a more complex experience in exchange for leveraging the existing methods.

It'd look something like
```fsharp
let output = StringBuilder()
runTestsWithCLIArgs [CLIArguments.OutputWriter (fun outputSegment -> output.Append(outputSegment) |> ignore)] [||] tests
output
```

To me, it seems unlikely we'll want any of the reflection-based runTests variants in interactive mode, but the complexity of
the method call can make a big difference in experience. Especially in an environment like FSI where you may have to type it over and over.

ratsclub:
Hmm, I was thinking of introducing a new case to `CLIArguments` along the lines of `LogOutput of LogOutput` and have it be handled inside `runTestsWithCLIArgsAndCancel`, this way your function `runTestsReturnLogs` would only pass a default configuration to `LogOutput`.

I mean, your code looks good to me, but I'm not convinced we should have a function for something that should be a configuration option and be documented on some sort of FAQ.

Expecto/Expecto.fs
ratsclub: What do you think about `runTestsInteractively`?

farlee2121: I thought of a similar name, but I feel like it denotes the test run itself is interactive, as in there will be decision points for the user during the test run.

farlee2121:
Suppose we have a call like `runTestsWithCLIArgsAndCancel [LogOutput someLogOutput] [||] tests`.
What is the someLogOutput doing? I don't think there's a consistent place we can redirect 
the log output for different environments, which means we either need to provide a variety of LogOutputs for different environments or the user defines their own (which kills the simplicity of an interactive experience). 

Do you have a more concrete vision of how LogOutput would work?

ratsclub:
> What is the someLogOutput doing?

I was considering something like this:

```fs
let runTestsForInteractive cliArgs args =
  let cliArgs = (LogOutput <some-cool-output>)::cliArgs

  runTestsWithCLIArgsAndCancel cliArgs args
```

This way we can control how the log output is handled inside `runTestsWithCLIArgsAndCancel`.

farlee2121:
The `some-cool-output` was the part I'm unclear about. 

In a polyglot notebook we could do something like `LogOutput (fun logs -> logs.Display())`.
But that won't work other places. 

If we accumulate in a string builder, then we're asking users to use at least 3 lines to display output, like demonstrated previously.

Expecto/Expecto.fs
ratsclub: Well, `runTestsForInteractive` seems reasonable to me, then.

ratsclub:
Oh, I ran an interactive session like on your screenshots and now I see what you mean. The *notebook* expects an explicit return value to output something, this is why we need to return a string itself... Well, in this case, I think this code does what is needed!

I'm still not comfortable expanding the API with this, but you perhaps found people that use this a lot. If it's useful for them, let's add it.

farlee2121:
Hmm, I can't claim high demand for interactive features. It's come up a few times (#365, #471) and it's something I periodically want to do. 

If we really don't want to add a method, I could pivot to the option like you suggested, then use it create my own 3rd-party package with the new method.

ratsclub:
> If we really don't want to add a method, I could pivot to the option like you suggested, then use it create my own 3rd-party package with the new method.

I don't want to block your work here, if you feel that we should add this, let's merge it!

I just feel that a third-party is a better way to deal with it, indeed. It just doesn't feel correct to introduce a new function on the library's API that is responsible to change something that should be a configuration.

farlee2121:
I guess I'll think about it.

farlee2121:
I added some configuration that would allow the interactive experience to be factored into an external library: `CLIArguments.LoggingConfigFactory`. 
It has to be a factory, or else there's no sensible way to get the configured verbosity.

I'm a bit concerned that this configuration looks like it's per-call but really it has to lean on `Logging.Global.initialize`.
Thus, all calls within a shared process will share the same LoggingConfig.


As for the interactive run endpoint, I do think it still belongs in the core library.
Our own readme talks about running in interactive environments ([here](https://github.com/haf/expecto#running-tests) and  [here](https://github.com/haf/expecto#filtering-with-filter))

PR #483, opened 2023-11-13
Add *testCaseTask functions

ratsclub:
This PR introduces the missing `*testCaseTask` functions.

For context, I wanted to remove code similar to this from the codebases I work with: https://github.com/demystifyfp/FsToolkit.ErrorHandling/blob/a0fa8595acd4c4581e8e0cb8586e45c7a7f5a47b/tests/FsToolkit.ErrorHandling.TaskResult.Tests

### DIFF
--- a/Expecto/Model.fs
+++ b/Expecto/Model.fs
@@ -2,6 +2,7 @@ namespace Expecto
 
 open System
 open System.Threading
+open System.Threading.Tasks
 
 type SourceLocation =
   { sourcePath: string; lineNumber: int }
--- a/Expecto/Test.fs
+++ b/Expecto/Test.fs
@@ -141,4 +141,3 @@ module Test =
     | Async test -> timeoutAsync test |> Async
     | AsyncFsCheck (testConfig, stressConfig, test) ->
       AsyncFsCheck (testConfig, stressConfig, test >> timeoutAsync)
-
--- a/Expecto.Tests/Tests.fs
+++ b/Expecto.Tests/Tests.fs
@@ -1396,6 +1396,20 @@ let taskTests =
         fun ms -> task { return ms.CanWrite ==? true }
     ]
 
+    testCaseTask "simple" <| task {
+      Expect.equal 1 1 "1=1"
+    }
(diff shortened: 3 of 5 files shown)


Expecto/Model.fs
```diff
@@ -53,6 +54,7 @@ type TestCode =
   | Sync of stest: (unit -> unit)
   | SyncWithCancel of stest: (CancellationToken -> unit)
   | Async of atest: Async<unit>
+  | Task of atest: Task<unit>
```
ratsclub, line 57: In order to avoid `TestCode.Task`, maybe call this `AsyncTask`?

Expecto/Test.fs
```diff
@@ -141,4 +141,5 @@ module Test =
     | Async test -> timeoutAsync test |> Async
     | AsyncFsCheck (testConfig, stressConfig, test) ->
       AsyncFsCheck (testConfig, stressConfig, test >> timeoutAsync)
+    | AsyncTask test -> timeoutAsync (test |> Async.AwaitTask) |> Async
```
farlee2121, line 144: Might we as well convert to async right away (i.e. in `testCaseTask`) instead of adding a new `TestCode` union case?
Is there a reason you chose the extra union case?

ratsclub: > Is there a reason you chose the extra union case?

I thought that it would be better to explicitly say the test is a task than to pass it as a converted `Async<'T>`, but now I think this is not useful.

farlee2121 (approved):

ratsclub:
@farlee2121 What's the process to release a new minor with this addition?

farlee2121:
There's a [FAKE target](https://github.com/haf/expecto/blob/main/CONTRIBUTING.md#releasing) that automates release.
You'll need to set environment variables for `GITHUB_TOKEN` and `NUGET_TOKEN`. The version number is taken from `RELEASE_NOTES.md`, so you'll need to update that too.

Might be best if we push our changes out in one release, assuming no issues come up with my PR

PR #484, opened 2023-11-20
Fix testCaseTask to accept a task, not an Async

farlee2121:
Fix testCaseTask to accept a task, not an Async. Also fix test name conflicts.

I noticed this was causing build troubles when I went back to the interactive PR, so I fixed it.

### DIFF
--- a/Expecto.Tests/PerformanceTests.fs
+++ b/Expecto.Tests/PerformanceTests.fs
@@ -24,24 +24,7 @@ let performance =
     testCase "md5 equals sha256" (fun _ ->
       Expect.isFasterThan runMD5 runSHA256 "MD5 equals SHA256 should fail"
     ) |> assertTestFailsWithMsgContaining "same"
-
-    testCase "sha256 versus md5" (fun _ ->
-      if RuntimeInformation.IsOSPlatform(OSPlatform.Linux) then
-        skiptest "Doesn't hold true on this platform for unclear reasons"
-      Expect.isFasterThan
-        (runSHA256 >> ignore |> repeat10)
-        (runMD5 >> ignore |> repeat10)
-        "SHA256 is faster than MD5 should fail"
-    ) |> assertTestFailsWithMsgContaining "slower"
-
-    testCase "md5 versus sha256" <| fun _ ->
-      if RuntimeInformation.IsOSPlatform(OSPlatform.Linux) then
-        skiptest "Doesn't hold true on this platform for unclear reasons"
-      Expect.isFasterThan
-        (runMD5 >> ignore |> repeat10)
-        (runSHA256 >> ignore |> repeat10)
-        "MD5 is faster than SHA256"
-    ]
+  ]
 
 [<Tests>]
 let findFastest =
--- a/Expecto.Tests/PerformanceTests.fs
+++ b/Expecto.Tests/PerformanceTests.fs
@@ -16,6 +16,7 @@ let runMD5() = md5.ComputeHash data
 let runSHA256() = sha256.ComputeHash data
 
 
+open System.Runtime.InteropServices;
 [<Tests>]
 let performance =
   testSequenced <| testList "performance cryptography tests" [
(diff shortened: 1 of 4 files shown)


Expecto.Tests/Tests.fs
```diff
@@ -1396,57 +1396,61 @@ let taskTests =
         fun ms -> task { return ms.CanWrite ==? true }
     ]
 
-    testCaseTask "simple" <| task {
```
farlee2121, line 1399: I didn't change any of the tests, just grouped them with testList to fix the name conflict

Expecto/Expecto.fs
```diff
@@ -88,23 +88,23 @@ module Tests =
   /// Builds an async test case
   let inline testCaseAsync name test = TestLabel(name, TestCase (Async test,Normal), Normal)
   /// Builds an async test case from a task
-  let inline testCaseTask name test = TestLabel(name, TestCase (Async test,Normal), Normal)
+  let inline testCaseTask name test = TestLabel(name, TestCase (Async (Async.AwaitTask test),Normal), Normal)
```
farlee2121, line 91: This was the main fix. `testCaseTask` and it's variants were expecting an Async instead of a Task

build.fsx
```diff
@@ -169,17 +169,17 @@ Target.create "Release" (fun _ ->
...
-// "Clean"
-//   ==> "BuildExpecto"
-//   ==> "BuildBenchmarkDotNet"
-//   ==> "BuildTest"
-//   ==> "RunTest"
-//   ==> "Pack"
-//   ==> "All"
-//   ==> "Push"
-//   ==> "Release"
+"CheckEnv"
```
farlee2121, line 172: I think this is why the build didn't fail it before

farlee2121:
For some reason these two tests fail on the build server
- performance cryptography tests.md5 versus sha256
- performance cryptography tests.sha256 versus md5

Perhaps some cross-platform nonsense going on here

farlee2121:
Not really sure why those cryptography tests exist, there are other tests that verify Expect.isFasterThan.

It doesn't seem like it should be, but the tests failing does seem to be a platform issue. I booted up a codespace and got the same result as the build server...

ratsclub:
Other than the small comment I made, this seems nice!

Expecto.Tests/PerformanceTests.fs
```diff
@@ -25,13 +26,17 @@ let performance =
...
+        skiptest "Doesn't hold true on this platform for unclear reasons"
       Expect.isFasterThan
         (runSHA256 >> ignore |> repeat10)
         (runMD5 >> ignore |> repeat10)
         "SHA256 is faster than MD5 should fail"
     ) |> assertTestFailsWithMsgContaining "slower"
 
     testCase "md5 versus sha256" <| fun _ ->
+      if RuntimeInformation.IsOSPlatform(OSPlatform.Linux) then
+        skiptest "Doesn't hold true on this platform for unclear reasons"
```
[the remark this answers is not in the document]
farlee2121: I was hesitant to remove them, since I don't quite understand the purpose of the tests.
But it also seems like the same functionality is already covered by [these tests](https://github.com/haf/expecto/blob/d9c19fbd1d0868ed81b4ab27143f65bd3bec00cb/Expecto.Tests/Tests.fs#L1453).

I think I'll just go ahead and remove them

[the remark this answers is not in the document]
farlee2121: Done

ratsclub (approved):

TheAngryByrd:
👋 I know this review is a little late but these should really take a `unit -> Task` or `unit -> Task<unit>`.  Problem with just using task, is they are `hot` meaning it will run immediately.  This will be a problem when using `p` on testLists and such.

farlee2121:
That's a very good point. Hmm. I'm not sure how we handle that since it's already out there.
It just went out hours ago though, so no one should have ingrained it into their code yet.

farlee2121:
@TheAngryByrd @ratsclub I whipped up a PR with a potential resolution, #492, if you'd be so kind as to provide thoughts

PR #489, opened 2024-03-13
Support ValueTask and Async bindings in testTask

### LINKED ISSUE #488: testTask doesn't handle the same bindings as task expressions
The F# `task` expression [can bind Tasks, ValueTasks, and Async types](https://learn.microsoft.com/en-us/dotnet/fsharp/language-reference/task-expressions#binding-by-using-let). It appears `testTask` currently only handles Task types.

Example

```fsharp
let iAmTask () : Task<int> = Task.FromResult(5)
let iAmValueTask () : ValueTask<int> = ValueTask.FromResult(5)
let iAmAsync () : Async<int> = async.Return 5

// Task expression works for all types
let example () = task {
    let! boundTask = iAmTask()
    let! boundValueTask = iAmValueTask()
    let! async = iAmAsync() 
    return ()
}

// testTask errors on ValueTask and Async
let example () = testTask "I am Test" {
    let! boundTask = iAmTask()
    let! boundValueTask = iAmValueTask()
    let! async = iAmAsync() 
    return ()
}
```

### Potential Solution

I'm still trying to understand the [TaskBuilder](https://github.com/dotnet/fsharp/blob/9669913821972b3d278b76f8b0006ba194827140/src/FSharp.Core/tasks.fs) code in F#, but I think we could get away with inheriting from TaskBuilder and overloading Run. All of the `TestTaskBuilder` methods except for run are a direct forwarding to `TaskBuilder`. This would allow us to stay in sync with whatever the task expression supports.

If inheriting doesn't work, we could probably get close enough with a few additional Bind overloads.

farlee2121:
Fixes #488 

Adds support for ValueTask<T>, ValueTask, and Async<T> bindings to the testTask expression.

This is accomplished through overloads of the Bind method.
I looked into inheriting the TaskBuilder, but it's constructor is hidden by the [tasks.fsi file](https://github.com/dotnet/fsharp/blob/9669913821972b3d278b76f8b0006ba194827140/src/FSharp.Core/tasks.fsi).

### DIFF
--- a/Expecto/Expecto.fs
+++ b/Expecto/Expecto.fs
@@ -243,6 +243,13 @@ module Tests =
       | Focused -> ftestCaseTask name a
       | Pending -> ptestCaseTask name a
 
+  [<AutoOpen>]
+  module TestTaskExtensions =
+    type TestTaskBuilder with
+      member inline __.Bind(p1:ValueTask<'a>, p2:'a->_) = task.Bind(p1, p2)
+      member inline __.Bind(p1:ValueTask, p2:unit->_) = task.Bind(p1, p2)
+      member inline __.Bind(p1:Async<'a>, p2:'a->_) = task.Bind(p1, p2)
+
   /// Builds a task test case
   let inline testTask name =
     TestTaskBuilder (name, Normal)
--- a/Expecto.Tests/Tests.fs
+++ b/Expecto.Tests/Tests.fs
@@ -1450,6 +1450,22 @@ let taskTests =
       testTask "inner skip" {
         skiptest "skipped"
       }
+      testTask "can let! bind ValueTask<T>" {
+        let expected = 5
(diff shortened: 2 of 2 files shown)


Expecto/Expecto.fs
```diff
@@ -243,6 +243,13 @@ module Tests =
       | Focused -> ftestCaseTask name a
       | Pending -> ptestCaseTask name a
 
+  [<AutoOpen>]
+  module TestTaskExtensions =
+    type TestTaskBuilder with
```
farlee2121, line 248: Adding these overloads as extensions helps the compiler know to prioritize the original Task overload, which prevents breakage in existing code that may have implicitly typed an expression based on Bind originally only taking variations of Task.

farlee2121:
Any thoughts from @ratsclub before I merge?

ratsclub (approved):

PR #492, opened 2024-03-15
Fix immediate task execution when defining test cases

farlee2121:
I.e. testTask and testCaseTask would start the task as soon as the test definition is created instead of when the test suite is run. This fixes a breaking change from how testTask used to work and a flaw of the new testCaseTask api. It is a breaking change to the testCaseTask api.

Thanks to @TheAngryByrd for [noticing this issue](https://github.com/haf/expecto/pull/484#issuecomment-2000168095) on pr
#484

### DIFF
--- a/Expecto.Tests/Tests.fs
+++ b/Expecto.Tests/Tests.fs
@@ -1397,19 +1397,23 @@ let taskTests =
     ]
 
     testList "testCaseTask" [
-      testCaseTask "simple" <| task {
-        Expect.equal 1 1 "1=1"
-      }
+      testCaseTask "simple" <| fun () ->
+        task {
+          Expect.equal 1 1 "1=1"
+        }
 
-      testCaseTask "let" <| task {
-        let! n = async { return 1 }
-        Expect.equal n 1 "n=1"
-      }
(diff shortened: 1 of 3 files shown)


farlee2121:
My thought is to release this as 10.2.1 and deprecate the 10.2.0 packages. 
This release isn't backwards compatible with 10.2.0, but it fixes a backwards compatibility issue introduced by 10.2.0 and only breaks compatibility with endpoints introduced in that same release.

TheAngryByrd (approved):
Thanks for this! Funny enough I was hanging out with @AngelMunoz in discord and he hit this :)

TheAngryByrd:
> My thought is to release this as 10.2.1 and deprecate the 10.2.0 packages. This release isn't backwards compatible with 10.2.0, but it fixes a backwards compatibility issue introduced by 10.2.0 and only breaks compatibility with endpoints introduced in that same release.
> 
> Thoughts?


Yeah I think that's a reasonable approach vs just bumping major.

farlee2121:
New version is out, 10.2.0 is deprecated with an explanatory message

TheAngryByrd:
Thank you!!

Issue #493, opened 2024-03-17
An observation / question about using testTheory with 'null' and '"" as inputs


Numpsy:
An observation whilst trying to do a few simple error case input tests with ```testTheory``` - 

I noticed that if I have test code like

```fsharp
[<Tests>]
let tests =
  testList "samples" [
    testTheory
        "null and empty arent the same"
        [ ""; null ]
        (fun value ->

            printf "%s" value)
  ]

```

The the test explorer in Ionide only showed one test:

![image](https://github.com/haf/expecto/assets/1178570/8d3870af-bf77-4400-bc47-9ea2b91ddd0b)

And then when I tried running from the command line, I got
```
[11:06:04 INF] Found duplicated test names, these names are: [["samples"; "null and empty arent the same"; ""]] <Expecto>
```

Maybe it's not a typical thing to do in F#, but for comparison If I do the same thing in NUnit with TestCase then it puts the string 'null' in the test name rather than an empty string so that the two tests show up differently:
![image](https://github.com/haf/expecto/assets/1178570/4d45e908-0fc9-4df4-a98a-47a18b412f8e)

So maybe it'd be possible to do something similar here?

Thanks

farlee2121:
Thanks for the examples!

It sounds like the problem is that null and empty string are, by default, rendered as the same string so the test name for both cases is the same and the test runner gets mad, thus displaying only one test.

I'd guess the solution here is to introduce a pretty printer that detects special values like this.

That'd probably be changed [here](https://github.com/haf/expecto/blob/cfa55e6c2ee1bed636406799ec09547656d2584b/Expecto/Expecto.fs#L128)
```fsharp
testCase (string case) <| fun () ->
//        ^^^^^^^
```

Just conjecture though. I haven't tried anything yet

Numpsy:
> Just conjecture though. I haven't tried anything yet

Just from a quick go with Expectos own tests, it does look like a different string function like
```fsharp
  let inline private stringify (value : 'T) =
      if typeof<'T> = typeof<string> then
        if value = Unchecked.defaultof<'T> then "null" else value.ToString()
      else
        string value

  /// Builds a theory test case
  let inline testTheory name cases test =
    let caseToTest case =
      testCase (stringify case) <| fun () ->
        test case |> ignore
    testList name (cases |> Seq.map caseToTest |> List.ofSeq)
```
lets these cases pass
```fsharp
    testTheory "odd numbers" [1; 3; 5] <| fun x ->
[...]

PR #494, opened 2024-03-17
Try to differentiate between null and empty strings in test theories

Numpsy:
refs #493 

Some testing to pin down where the issue is - not sure if this is the best apporach to fixing it, or if there are other types than strings that might have similar issues.

### DIFF
--- a/Expecto.Tests/Tests.fs
+++ b/Expecto.Tests/Tests.fs
@@ -1834,11 +1834,22 @@ let theory =
     testTheory "odd numbers" [1; 3; 5] <| fun x ->
       Expect.isTrue (x % 2 = 1) "should be odd"
 
+    testTheory "empty strings" [""; null] <| fun x ->
+      Expect.isTrue (System.String.IsNullOrEmpty(x)) "should be null or empty"
+
     testTheoryAsync "async even numbers" [2; 4; 6] <| fun x -> async {
       Expect.isTrue (x % 2 = 0) "should be even"
     }
 
+    testTheoryAsync "async empty strings" [""; null] <| fun x -> async {
+      Expect.isTrue (System.String.IsNullOrEmpty(x)) "should be null or empty"
+    }
+
     testTheoryTask "task odd numbers" [1; 3; 5;] <| fun x -> task {
       Expect.isTrue (x % 2 = 1) "should be odd"
+    }
+
+    testTheoryTask "task empty strings" [""; null] <| fun x -> task {
+      Expect.isTrue (System.String.IsNullOrEmpty(x)) "should be null or empty"
     }
   ]
--- a/Expecto.Tests/Tests.fs
+++ b/Expecto.Tests/Tests.fs
@@ -1834,22 +1834,24 @@ let theory =
     testTheory "odd numbers" [1; 3; 5] <| fun x ->
       Expect.isTrue (x % 2 = 1) "should be odd"
 
-    testTheory "empty strings" [""; null] <| fun x ->
-      Expect.isTrue (System.String.IsNullOrEmpty(x)) "should be null or empty"
-
     testTheoryAsync "async even numbers" [2; 4; 6] <| fun x -> async {
       Expect.isTrue (x % 2 = 0) "should be even"
-    }
-
-    testTheoryAsync "async empty strings" [""; null] <| fun x -> async {
-      Expect.isTrue (System.String.IsNullOrEmpty(x)) "should be null or empty"
     }
 
     testTheoryTask "task odd numbers" [1; 3; 5;] <| fun x -> task {
       Expect.isTrue (x % 2 = 1) "should be odd"
     }
 
-    testTheoryTask "task empty strings" [""; null] <| fun x -> task {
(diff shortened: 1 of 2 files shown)


farlee2121:
Overall, looks good. Gives us a place to make changes if we find any more special values that need differentiated. 

I made one change: I grouped the new tests in an explanatory test list.

Checking off a few other potential special values
- white space should produce different values
- numeric types... I can't think of any issues
- weird chars... zero-width characters? 
- reference types... this could be a problem. Classes don't have good default string representations

Example of breaking reference type
```fsharp
type ClassA(i: int) = 
  let i = i

testTheory "I am theory" [ClassA(5); ClassA(4)] (fun (refType: ClassA) -> true)
```

I'm not sure how common this reference type scenario is. The empty vs null one seems quite common though, and it's worth getting a fix in for that at least.

farlee2121:
Also, thanks for the PR!

Numpsy:
> * white space should produce different values

With the current change, it works to run the tests, but the Visual Studio test explorer looks a bit 'unfriendly' - 
image

If I borrow the quoting approach from NUnit
```fsharp
if value = Unchecked.defaultof<'T> then "null" else $"\"{value}\""
```

then it looks like this, which seems nicer:

image
image

Numpsy:
> weird chars... zero-width characters?

I don't know if there are any potential locale specific issues that might occur?

> reference types... this could be a problem. Classes don't have good default string representations

I can't recall offhand what NUnit/XUnit do for that, will have to test.

farlee2121:
Using NUnit and XUnit theories as a reference is a great idea

Numpsy:
Ok, I don't think you can specify class instances in the NUnit ```TestCase``` attribute, and if I use ```TestCaseSource``` then I get this, where it only lists one test but then has two results, and the display name in the test explorer is just the class type - 

Visual Studio:
image
 
Ionide:
image

farlee2121:
C# attributes can only accept constant parameters. Makes sense that class instances are out.

That's interesting that TestCaseSource behaves differently than the attributes. 
That still leaves us with an open question on how to handle classes.

Did you check out xUnit too?

Numpsy:
just tried adding a 'TheoryData' based test to an Xunit based test project and got

image

So it's still showing one test with several results, but in that case it's managing to show a string representation of the class instance instead of just showing the type name

farlee2121:
Thanks for the research!

Maybe we could mimic the xUnit approach to class rendering. That could open new issues with special characters in test names though...
We could potentially detect duplicate test names and merge them into a single test with the cases in the description, but that feels like an unintuitive user experience.

I think it's worth merging the string/null fix by itself, since that's a fairly common issue.
It's less clear how common the class issue would be. I also doubled checked that records and unions have sensible default string representations and wouldn't suffer from the same issue as classes.

Numpsy:
> It's less clear how common the class issue would be.

I don't recall having used classes with TestCase myself (it's usually base types like strings, numbers, enums). so making the string case work fixes my problem.

> That could open new issues with special characters in test names though
> We could potentially detect duplicate test names and merge them into a single test with the cases in the description

I'm not sure if putting that data in the test names would have issues if the class instances had dynamic field in them? (e.g. say you had a test class instance with a time stamp or guid or something in it, trying to render that as part of the test name rather than just in the results might be problematic?)

farlee2121 (approved):

farlee2121:
The characters that cause issues (that I'm aware of) are `+` and `.`, which are used as separators for test segments.
I don't think guids or dates would typically contain those.

I've also seen some filter issues with characters like `()&|`, but it's on the test explorer to escape those kinds of characters.

farlee2121:
Oof. Glad I thought to test before pushing the packages.
It looks like quotes around strings causes issues with the Ionide test explorer. I'll see if it can be fixed

farlee2121:
A fix is PRed https://github.com/ionide/ionide-vscode-fsharp/pull/1999

Numpsy:
> Oof. Glad I thought to test before pushing the packages. It looks like quotes around strings causes issues with the Ionide test explorer. I'll see if it can be fixed

Sorry, thought I'd tested that when I made the screen shots, must not have though :-(

My comment about dates etc was more about 'dynamic' data, as in what will the test explorer do if the test name were made of data that might be different between runs (maybe not common, but could happen with class types?) - although maybe you could have a ```+``` in the string for a time value as an offset from UTC.

Your comment about ```.``` made me think about floating point numbers as well though - which currently look like this in Ionide:

image

Numpsy:
> * reference types... this could be a problem. Classes don't have good default string representations

Just accidentally fell over a related situation here - If I try to give a testTheory arrays of strings as inputs, then each array has the same string representation so you get a ```Found duplicated test names``` error again

farlee2121:
While not very elegant, a workaround could be to give each case a data point to differentiate it. I.e. `[("case1", [array here]); ("case2", [other array])`.

Actually, we could potentially generalize such an approach.  Automatically add a case identifier to make sure each case is unique even if the arguments don't have distinct string representations.

PR #497, opened 2024-04-11
WIP:  Opentelemetry

TheAngryByrd:
I used the [Aspire](https://learn.microsoft.com/en-us/samples/dotnet/aspire-samples/aspire-standalone-dashboard/) dashboard. You can run it via docker:

```bash
docker run --rm -it -p 18888:18888 -p 4317:18889 -d --name aspire-dashboard mcr.microsoft.com/dotnet/nightly/aspire-dashboard:8.0.0-preview.4
```

Then you can run the `Expecto.Tests` project. It should export OpenTelemetry traces to Aspire.

Example of traces for tests showing up:

image

Clicking View on one of them with a red dot (red dot indicates there were exceptions/errors):

image

Clicking on the top span, you can see details about its run.

image

Clicking on one of the child spans you can see details such as Exception data.

image

image

I didn't show it here but you can also export logs that are associated with a test trace.


Expecto/Expecto.Impl.fs
```diff
@@ -9,6 +10,20 @@ open Expecto.Logging.Message
 open Helpers
 open Mono.Cecil
 
+//! The other option is to use a dedicated activity source for Expecto instead of adding it to the config
```
TheAngryByrd, line 13: One option is to have Expecto have a dedicate ActivitySource, the other is to add it the the Expecto config and allow a user to pass it along.

Expecto/Expecto.Impl.fs
```diff
@@ -520,6 +535,11 @@ module Impl =
       colour: ColourLevel
       /// Split test names by `.` or `/`
       joinWith: JoinWith
+      // One option is to allow the consumer to provide an activity source
+      // only problem is the only way to update the config is by using the CLIArguments currently
+      // we would have to add a new CLIArgument but that doesn't really work as it's not a reallyCLI option
+      // or have another way of updating the config after it's been created
+      activitySource : ActivitySource option
```
TheAngryByrd, line 542: As mentioned, we can have an activity source in the config but there's not really a function that lets people update the config.

It might be more tricky to do this option if people are using the `dotnet test` integration so maybe a dedicated source is appropriate.

Expecto/Expecto.Impl.fs
```diff
@@ -559,8 +580,63 @@ module Impl =
               }
       }
 
+  let inline internal setStatus (status : ActivityStatusCode) (span : Activity) =
```
TheAngryByrd, line 583: Lots of helper functions. Don't have to live here, just here for the moment.

Expecto/Expecto.Impl.fs
```diff
@@ -559,8 +580,63 @@ module Impl =
...
+  let inline internal createActivity (name : string) (source : ActivitySource option) =
+    match source with
+    | Some source when not(isNull source) -> source.CreateActivity(name, ActivityKind.Internal)
+    | _ -> null
+
   let execTestAsync (ct:CancellationToken) config (test:FlatTest) : Async<TestSummary> =
     async {
+      let span =
+        config.activitySource
+        |> createActivity (config.joinWith.format test.name)
```
TheAngryByrd, line 636: This is the starting point for creating an span. Later we add additional information to it based on whether we pass/fail/ignore.

Expecto.Tests/Main.fs
```diff
@@ -2,13 +2,42 @@ module Main
...
 let main args =
+  let activitySource = new ActivitySource(serviceName)
+  use traceProvider =
+          Sdk
+              .CreateTracerProviderBuilder()
+              .AddSource(serviceName)
+              .SetResourceBuilder(resourceBuilder )
+              .AddOtlpExporter()
+              .Build()
+  let tracer = traceProvider.GetTracer(serviceName)
```
TheAngryByrd, line 33: This is adding the OpenTelemetry Exporter to the test project, mostly for demo purposes. Could create a separate project if desired.

Expecto.Tests/Main.fs
```diff
@@ -2,13 +2,42 @@ module Main
...
+  // use span = tracer.StartActiveSpan("Expecto.main")
+  use span = tracer.StartRootSpan("Expecto.main")
   let test =
     Impl.testFromThisAssembly()
     |> Option.orDefault (TestList ([], Normal))
     |> Test.shuffle "."
-  runTestsWithCLIArgs [NUnit_Summary "bin/Expecto.Tests.TestResults.xml"] args test
+  runTestsWithCLIArgs [NUnit_Summary "bin/Expecto.Tests.TestResults.xml"; CLIArguments.ActivitySource activitySource] args test
```
TheAngryByrd, line 40: Here a consumer would pass in the ActivitySource. Otherwise they'd have to add the dedicated Expecto one to the `.AddSource` call on line 29.

farlee2121:
I'm still thinking over generalization approaches. I haven't found a clear nice answer yet.

Have you considered how the current span approach would behave with stress testing?

TheAngryByrd:
Oh one of the things missing too is how to integrate with `dotnet test` since we can't call `addOpenTelemetry_SpanPerTest`, we'd probably need some attribute that would do the call on our behalf when calling into `testFromAssembly`

farlee2121:
Upon review, I think I got wrap up in my own line of thought and didn't interpret your statement correctly.

I'm not sure I see the incompatibility with `dotnet test`. Doesn't YoloDev.Expecto.TestSdk respect whatever is configured in main?

TheAngryByrd:
As far as I know, dotnet test doesn't care whats in your main/entrypoint. Yolodev uses [Expecto.Impl.testFromAssembly](https://github.com/YoloDev/YoloDev.Expecto.TestSdk/blob/57a63a2db6b100ab226aa42bf42491b41cce965d/src/YoloDev.Expecto.TestSdk/discovery.fs#L64) to find tests.

I suppose you can call `addOpenTelemetry_SpanPerTest` for every `testlist` that uses `[<Tests>]`. There _might_ be some tricks required to get `ActivitySource` to not be null, I'll have to look later.

farlee2121:
Interesting. I hadn't thought of such a limitation.

Tests names can be listed via dotnet run, but that doesn't provide fully-hydrated Test records.
I don't see a clear solution to this either, since I'm not sure anonymous functions could be effectively transferred out-of-process.

TheAngryByrd:
So creating the TraceProvider statically:

```fsharp
do
    let provider = traceProvider()
    AppDomain.CurrentDomain.ProcessExit.Add(fun _ -> provider.Dispose())
```

And `addOpenTelemetry_SpanPerTest` has to be added per `[<Tests>]`. Which isn't too bad but might get annoying for people that have lots of `[<Tests>]` instead of a few.

After that, this does work for `dotnet test`

TheAngryByrd:
👋 Long time no push.

I'm thinking I'd want to try to get this into Expecto. What do you think would be the acceptance for this? Somethings come to mind:

- Review Public interface
- Docs showing how to do this
- Tests actually using OTel during CI

farlee2121:
I scanned over the code again. It looks like the essence of the change is concentrated in OpenTelemetry.fs and a test. 

A few items we'd want to address
- As you mentioned, tests and docs would be good.
- It looks like there's some left over `open` statements in various files from previous iterations that could be trimmed.
- `addOpenTelemetry_SpanPerTest` takes an `ExpectoConfig`, which isn't part of the public API anymore. Following the current public API, it'd take a sequence of `CLIArguments`
- The code needs moved out of the Expecto.Tests project. Preferably into its own project/package.
  - The tests need split out too
  - the project needs to be added to various FAKE targets (build, test, pack, "push"/pushing packages to nuget)

PR #501, opened 2024-08-09
Fix unusable replay seed with FsCheck3

rynoV:
When using FsCheck3 it was impossible to use the replay seeds given in the failed test output because the config values only accepted `int32`, instead of `uint64`

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,6 @@
+### 11.0.0 - 2024-08-09
+* Fix unusable replay seed with FsCheck3
+
 ### 10.2.1 - 2024-03-15
 * Fix bug where testTask and testCaseTask allow the tasks to start immediately when the test is defined, breaking backward compatibility with testTask.
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,8 @@
-### 11.0.0 - 2024-08-09
-* Fix unusable replay seed with FsCheck3
+### 11.0.0 - 2024-08-12
+* Breaking Change: move `FsCheckConfig.replay` from `int` to `uint64`.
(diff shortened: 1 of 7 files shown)


RELEASE_NOTES.md
```diff
@@ -1,3 +1,6 @@
+### 11.0.0 - 2024-08-09
+* Fix unusable replay seed with FsCheck3
```
farlee2121, line 2: Since this is a breaking change, the release notes should be more descriptive of the breaking change.
Perhaps something like 
```
Breaking Change: move FsCheckConfig.replay from int to uint64. 
Fixes issue where many FsCheck3 runs could not be replayed since the random seed is too large. 
Existing FsCheck 2 users should be able to use the same seeds values, but converted to uint64. 
Uint64 literals can be defined like `let iAm5 = 5UL` 
```

farlee2121:
Thanks for the PR. Definitely a shortcoming of the original solution.

I notice that FsCheck seems to be pushing toward releasing version 3.
> but 3.x is the version that is currently maintained. Only v3 will get new features and fixes, even though it's in pre-release.

Perhaps it's time to make fscheck 3 the main Expecto.FsCheck?

RELEASE_NOTES.md
farlee2121: Oh. The ReadMe also needs updated with the new replay type. Probably also with update guidance

rynoV: Updated thanks, let me know if there's anything I missed

rynoV:
> Perhaps it's time to make fscheck 3 the main Expecto.FsCheck?

This seems reasonable to me, I don't use fscheck 2

farlee2121:
Thanks for the updates. I haven't looked at them yet, but I'll try to do it this week.

farlee2121:
11.0.0-alpha1 is out

Issue #495, opened 2024-03-29
Please specify what versions of Benchmark.net are compatible with the current Expecto release in the README


sonicbhoc:
BenchmarkDotNet's `IConfig` interface has changed multiple times over the years, and the latest version pulled in by nuget (0.13.12) is incompatible with expecto. The error message I get is:

`System.TypeLoadException: Method 'GetEventProcessors' in type 'BenchmarkConfig' from assembly 'Expecto.BenchmarkDotNet, Version=10.1.0.0, Culture=neutral, PublicKeyToken=null' does not have an implementation.`

Turns out that this is a new field added in BenchmarkDotNet v0.13.9.

I would suggest clearly mentioning somewhere visible (like the README) what version of BenchmarkDotNet each version of Expecto is tested against just to make it clear what version of the library works.

farlee2121:
I could get that reviewed and merged real quick if you'd like to draft up a change.

sonicbhoc:
I can take a crack at it soon. I am a little swamped right now though.

Numpsy:
On the subject of BenchmarkDotNet - there is a new 0.14.0 release which has some useful new features / fixes - are there any thoughts on updating to that if there are breaking API changes within the 0.13.X set anyway?

PR #502, opened 2024-08-14
Update BenchmarkDotNet to 0.14.0

Numpsy:
refs #495 

Updates BenchmarkDotNet  to the latest 0.14.0 release, and extends the Expecto code to implement the new members of ```IConfig``` - no changes other than that.

The changes to the properties here probably make this count as a breaking API change, but if  #501 is already a breaking change then perhaps that's an opportunity to do both at once?

(Note: I won't change the release notes etc here yet, as it'd probably just clash with #501)


farlee2121:
Just merged #501, but I'll probably release it as a preview version first, since it's a breaking change 

This PR makes sense to me conceptually, but I'll admit that I don't really know much about Expecto.BenchmarkDotNet.

I scanned the [BenchmarkDotNet 14.0 release notes](https://benchmarkdotnet.org/changelog/v0.14.0.html) and our code. I didn't see any red flags. Is there anything else you were wanting to investigate before making this a full PR?

Numpsy:
> Is there anything else you were wanting to investigate before making this a full PR?

I'll have a look at adding a release note entry later, then activate it.

Numpsy:
> I don't really know much about Expecto.BenchmarkDotNet.

I want to try adding some new benchmarks to an existing Expecto test project, but I'm using ```UseArtifactsOutput``` in the build and BenchmarkDotNet 0.13 doesn't work with that - It's hopefully pretty straight forward, but I'll give it some more tests with a test or prerelease build

Numpsy:
Does this need an entry in ```Migration notes``` as well as the release note? (anyone who is creating their own instances of ```BenchmarkConfig``` instead of using the default would have to populate the two new members)

farlee2121:
Hmm. The changes seem pretty self explanatory. It's not nearly as disruptive as the fscheck 2 to 3 migration. The compiler warnings will likely tell people what they need to know.

farlee2121:
This should be out with 11.0.0-alpha2

PR #503, opened 2024-08-14
Update the fake-cli and Paket tools

Numpsy:
When running the build, I got a number of 
```
Could not detect any platforms from 'net8.0' in XXXX
```
warnings in the build.
(There are a few of them in the Ci, e.g. https://github.com/haf/expecto/actions/runs/10391743404/job/28775360903#step:4:257)

Those can happen because Paket 7 doesn't know what .NET 8 is.

So - any thoughts about updating it to version 8 to try to fix the warnings?

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,13 +3,13 @@
   "isRoot": true,
   "tools": {
     "paket": {
-      "version": "7.1.5",
+      "version": "8.0.3",
       "commands": [
         "paket"
       ]
     },
     "fake-cli": {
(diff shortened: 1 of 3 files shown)


farlee2121:
Looks good to me, and should be low risk.
Was there anything else you wanted to cover before making this a full PR?

farlee2121:
Also, thanks for the PRs!

Numpsy:
> Looks good to me, and should be low risk. Was there anything else you wanted to cover before making this a full PR?

Nope, was just checking the idea was ok.

PR #506, opened 2024-08-16
Fix the build for Expecto.BenchmarkDotNet.Tests

Numpsy:
I hadn't realised before, but it looks like the Expecto.BenchmarkDotNet.Tests test project isn't actually being built by the CI build - there are warnings in the build about not finding any projects to restore, e.g.

https://github.com/haf/expecto/actions/runs/10412005602/job/28836955136#step:4:355

And running it directly fails with an error about ```runTestsInAssembly``` not being found.

I don't know about the history of this project (maybe you don't actually want to run benchmarks during the CI  build?) - but it seems wrong to not at least compile it?

### DIFF
--- a/Expecto.BenchmarkDotNet.Tests/BenchmarkDotNetTests.fs
+++ b/Expecto.BenchmarkDotNet.Tests/BenchmarkDotNetTests.fs
@@ -43,4 +43,4 @@ let tests =
 
 [<EntryPoint>]
 let main args =
-    runTestsInAssembly defaultConfig args
+    runTestsInAssemblyWithCLIArgs [] args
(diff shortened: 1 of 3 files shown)


farlee2121:
Interesting. I agree that we'd at least want it building. 
Depending on how slow the tests are, we could just run them with the other tests or we could create a new step for tests that only run in the CI build

Numpsy:
I just recalled that something similar happened in FSharpLint where a BenchmarkDotNet project was silently broken due to not being compiled in the CI builds - https://github.com/fsprojects/FSharpLint/issues/620 - so best to build things to avoid that situation.

PR #507, opened 2024-08-18
Run Expecto.BenchmarkDotNet.Tests during the build

Numpsy:
Refs https://github.com/haf/expecto/pull/506#issuecomment-2295282351

The extra tests don't take very long on my laptop, so lets see how the CI does

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -73,6 +73,13 @@ let build project =
              Common = DotNet.Options.withDotNetCliPath dotnetExePath p.Common })
     project
 
+let runTest project =
+  Trace.logfn "Running %s on .NET Core" project
+  DotNet.exec (DotNet.Options.withDotNetCliPath dotnetExePath)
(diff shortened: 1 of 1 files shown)


farlee2121:
Looks like it added about a minute to the build. That's a fairly high percentage increase, but I'd say still definitely in the realm of a running on every CI build.
At the same time, I'd say it's probably worth separating this out of the default local build

Numpsy:
I've moved the BDN tests into a separate FAKE target called ```RunBenchmarkDotNetTest``` so that the regular unit tests can be run without running those, but as it stands the default target for local builds is 'All' so it still runs them by default locally

farlee2121:
Ah. I didn't think about the dependency knot in target order. It'd be good to have it run before pack, but pack is part of the local build.

I'll go ahead an merge it for now. We can re-think the local build if it becomes a burden

PR #509, opened 2024-08-23
A few fsharp-analyzers suggestions

Numpsy:
I tried running the Expecto library code through a few fsharp-analyzers, and got a few suggested changes.
These ones are from https://g-research.github.io/fsharp-analyzers/analyzers/VirtualCallAnalyzer.html where it suggests replacing Seq functions with explicit types when the collection types are fixed:

```

Expecto.Impl.fs(963,11): Warning GRA-VIRTUALCALL-001 : Consider replacing the call of Seq.tryFind with a function from the Array module to avoid the costs of virtual calls.
Expecto.Impl.fs(972,47): Warning GRA-VIRTUALCALL-001 : Consider replacing the call of Seq.find with a function from the Array module to avoid the costs of virtual calls.
Expecto.fs(546,12): Warning GRA-VIRTUALCALL-001 : Consider replacing the call of Seq.fold with a function from the List module to avoid the costs of virtual calls.
```

Any thoughts?

### DIFF
--- a/Expecto/Expecto.Impl.fs
+++ b/Expecto/Expecto.Impl.fs
@@ -960,7 +960,7 @@ module Impl =
     else
       let nestedFunc =
         t.GetTypeInfo().GetFields()
-        |> Seq.tryFind (fun f -> isFsharpFuncType f.FieldType)
+        |> Array.tryFind (fun f -> isFsharpFuncType f.FieldType)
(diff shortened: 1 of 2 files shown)


farlee2121:
I don't see any issue with this. It's not affecting any type inference.

PR #511, opened 2024-09-10
Add option to skip printing skipped tests + fix FsCheck3 replay

### LINKED ISSUE #453: Option to suppress "Skipped" messages
This would be particularly useful when focusing a test, or list of tests.

We have hundreds of tests, so focusing results in a flood of messages.

rynoV:
Resolves #453 : https://github.com/haf/expecto/pull/511/commits/79f6e21881ae566036a22bbf0f1879454bf5784a

Also fixes the replay config so fscheck 3 will skip directly to the failing case: https://github.com/haf/expecto/pull/511/files/79f6e21881ae566036a22bbf0f1879454bf5784a..701746a1b3a2773ab9a011b221ce15e2f9345d0e

### DIFF
--- a/Expecto.FsCheck/FsCheck.fs
+++ b/Expecto.FsCheck/FsCheck.fs
@@ -71,7 +71,7 @@ module ExpectoFsCheck =
                           (String.concat " " data.Labels)
 
               let focus =
-                sprintf "Focus on error:\n\t%s (%A, %A) \"%s\"" methodName (uint64 std) (uint64 gen) name
+                sprintf "Focus on error:\n\t%s (%A, %A, 0) \"%s\"" methodName (uint64 std) (uint64 gen) name
 
               sprintf "Failed after %s. %s%s\nResult:\n\t%A\n%s%s%s"
                       (numTests data.NumberOfTests) parameters shrunk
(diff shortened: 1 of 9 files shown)


farlee2121:
I probably won't be able to review this for a few weeks.

farlee2121:
Alright. I'm back and starting to give this a look.

I think these are distinct work items that should be considered separately. Trying to review them together muddies the picture of what's going on with either

farlee2121:
Could you explain the perceived fix to FsCheck replay?

rynoV:
Yeah my bad should have made two PRs. I updated the description with links to the diffs for the two changes

> Could you explain the perceived fix to FsCheck replay?

In FsCheck3 they seem to have added a size parameter for replay in addition to the seed. Without the size, replay was playing all test inputs leading up to the failure; this change allows skipping directly to the failing test input

farlee2121:
That makes sense. Skipping right to the broken case would be good for debugging.

I noticed that you also changed from printing the original seed to the final seed.
I'm unsure how that impacts users focused more on deterministically repeating a specific run. 
Expecto itself has some tests that use the seed that way

rynoV:
Ah didn't think about that. I added it to the output

farlee2121:
This fell off my radar.

I would need these work items to be split before merging either. 
A clean is history for future context is very important for a community maintained package like this.

From what I've see so far
- The Replay changes seem good 
- Not printing skipped test messages has me a bit more worried. There's already some challenging interactions/interdependence between the config and the printers. I'll need to consider it and alternative approaches more carefully

PR #515, opened 2025-01-06
Fix fscheck3 replay

rynoV:
Fixes the replay config to use the size parameter so fscheck 3 will skip directly to the failing case. Also updates test output to show the "final seed" which will directly run the failing case when used, instead of the "original seed"

Separated out from #511

### DIFF
--- a/Expecto.FsCheck/FsCheck.fs
+++ b/Expecto.FsCheck/FsCheck.fs
@@ -71,7 +71,7 @@ module ExpectoFsCheck =
                           (String.concat " " data.Labels)
 
               let focus =
-                sprintf "Focus on error:\n\t%s (%A, %A) \"%s\"" methodName (uint64 std) (uint64 gen) name
+                sprintf "Focus on error:\n\t%s (%A, %A, 0) \"%s\"" methodName (uint64 std) (uint64 gen) name
 
               sprintf "Failed after %s. %s%s\nResult:\n\t%A\n%s%s%s"
                       (numTests data.NumberOfTests) parameters shrunk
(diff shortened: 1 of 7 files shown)


Expecto.FsCheck/FsCheck.fs
```diff
@@ -71,7 +71,7 @@ module ExpectoFsCheck =
                           (String.concat " " data.Labels)
 
               let focus =
-                sprintf "Focus on error:\n\t%s (%A, %A) \"%s\"" methodName (uint64 std) (uint64 gen) name
+                sprintf "Focus on error:\n\t%s (%A, %A, 0) \"%s\"" methodName (uint64 std) (uint64 gen) name
```
farlee2121, line 74: What is the motivation for always printing zero here?

farlee2121:
Thanks for splitting these up!

This looks pretty good, I just had two questions that I left as comments.

Expecto.FsCheck/FsCheck.fs
rynoV: This is so the user can copy and paste. The FsCheck 2 module will just ignore the third value

Expecto/Model.fs
```diff
@@ -16,7 +16,7 @@ type FsCheckConfig =
     /// The size to use for the last test, when all the tests are passing. The size increases linearly between Start- and EndSize.
     endSize: int
     /// If set, the seed to use to start testing. Allows reproduction of previous runs.
-    replay: (uint64 * uint64) option
+    replay: (uint64 * uint64 * int) option
```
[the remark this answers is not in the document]
rynoV: Hmm I suppose not. I'll update to use an option, and update the fscheck2 stuff to use None

farlee2121:
Looks great. I'll get a preview package released soon.
Thanks for the contribution!

farlee2121:
The alpha4 package is up

PR #525, opened 2025-11-28
float32 and UoM assertion functions

jwosty:
I've added float32 versions of all the float assertion functions. I've also extended both the float and float32 versions of all of those to support floats/float32s with units of measure.

List of functions added:
* `floatClosef`
* `floatLessThanOrClosef`
* `floatGreaterThanOrClosef`
* `isNotNaNf`
* `isNotPositiveInfinityf`
* `isNotNegativeInfinityf`
* `isNotInfinityf`

For example we can now write:

```fsharp
Expect.floatClose Accuracy.low<m> 1.0<m> 2.0<m> "result"
```

and:

```fsharp
Expect.floatClosef Accuracy.lowf<m> 1.0f<m> 2.0f<m> "result"
```


jwosty:
Great - just pushed some changes to split `Accuracy` into `Accuracy` and `AccuracyF`, and rename `floatClosef` and similar to `float32Close`.

However in doing this I noticed there's still `isNaNf` - I think leaving this one makes sense, right? Since there's there's no 'float' word to hang it off of, and the 32 bit NaN literal is called `nanf` in F#. So someone searching for those would probably expect it to be called `isNotNaNf`. Same applies for `isInfinityf` and friends (literal is `infinityf`).

I also noticed there's no `isNaN`, `isPositiveInfinity`, `isNegativeInfinity`, and `isInfinity`. I've taken the liberty to add those. It's easy enough to revert that commit if you don't want them.

Overload experiment branch is here: https://github.com/jwosty/expecto/tree/float-overloads-srtp

jwosty:
@farlee2121 If it were my library, I'd probably do the operator overloads. But it is not my library so it's ultimately up to you and can understand wanting to not do the overloads if that's the choice you make

jwosty:
Sure thing. Well, since you're asking :) - my vote would be for the overloads, as reducing noisiness and increasing ergonomics for test code is a big deal IMO. Same reason why other test frameworks Assert classes often heavily rely on overloading. I personally think that, as long as we pay special attention to docs around this part, it outweighs the downside in slightly increased barrier to entry.

jwosty:
@farlee2121 thoughts? I'd love to get some form of this merged

farlee2121:
I'm planning to push Expecto 11 to general availability. 
Any more thoughts on this before I do?

jwosty:
@farlee2121 whoops, I missed your comment! I'm happy for you to merge this as-is whenever you'd like

farlee2121:
Should be live with version 11.1.0

Issue #498, opened 2024-07-20
Build expecto test apps as NativeAOT


Numpsy:
I'm not sure  how sensible / feasible doing this is, but:

If I try to build the minimal test app from the readme:
```fsharp
open Expecto

let tests =
  test "A simple test" {
    let subject = "Hello World"
    Expect.equal subject "Hello World" "The strings should equal"
  }

[<EntryPoint>]
let main args =
  runTestsWithCLIArgs [] args tests
```

as .NET 8.0 using NativeAOT, then the test app fails to run with this exception:
```
Unhandled Exception: System.TypeInitializationException: A type initializer threw an exception. To determine which type, inspect the InnerException's StackTrace property.
 ---> System.TypeLoadException: Attempted to load a type that was not created during ahead of time compilation.
   at Internal.Runtime.CompilerHelpers.ThrowHelpers.ThrowUnavailableType() + 0x2b
   at Microsoft.FSharp.Core.FSharpFunc`2.InvokeFast[V](FSharpFunc`2, T, TResult) + 0x1a
   at <StartupCode$Expecto>.$Logging..cctor() + 0x4b3
   at System.Runtime.CompilerServices.ClassConstructorRunner.EnsureClassConstructorRun(StaticClassConstructionContext*) + 0xb9
   --- End of inner exception stack trace ---
   at System.Runtime.CompilerServices.ClassConstructorRunner.EnsureClassConstructorRun(StaticClassConstructionContext*) + 0x14a
   at System.Runtime.CompilerServices.ClassConstructorRunner.CheckStaticClassConstructionReturnGCStaticBase(StaticClassConstructionContext*, Object) + 0xd
   at Expecto.Logging.ANSIOutputWriter.setColourLevel(ColourLevel) + 0x63
   at Expecto.Tests.runTestsWithCancel@600(CancellationToken, Impl.ExpectoConfig, Test) + 0x20
   at Expecto.Tests.runTestsWithCLIArgs(IEnumerable`1, String[], Test) + 0x1d
   at ex_test_2!<BaseAddress>+0x2268a0
```
(I've seen a few variations of the call stack with other test apps, depending on if they use runTestsInAssembly or variations like that).

If I build the app with full AOT analysis enabled then I get a few instances of warnings like this:
```
[...]

PR #530, opened 2026-04-05
Make Expecto libraries target netstandard2.0

roboz0r:
I did a review of the code and couldn't find anything that would prevent the library from targeting .NET Standard 2.0. I have a .NET Framework project that I'd like to be able to upgrade to the latest version of Expecto. The test projects can't target .NET Standard so I made them dual target `net6.0` and `net481`.


roboz0r:
I took a few more attempts at this, seems like it's now failing because it now needs both the .NET 10 and .NET 6 SDKs. Happy to keep attempting fixes on this if retargeting Expecto to .NET Standard 2.0 will be approved in principle.

farlee2121:
closes #524 

I think targeting netstandard2.0 would make sense, but I at least want to ask the person who changed it why they did so. I poked around and [the change](https://github.com/haf/expecto/commit/7071bbf353760c923bfd3342188b4ba0c0e7afd3) seems to have been @ratsclub 

@ratsclub Do you remember why expecto moved from netstandard2.0 to net6.0?

farlee2121:
I think good faith efforts have been made to understand previous changes and no clear reason has been uncovered.

I'd be comfortable going forward with returning Expecto to netstandard2.0.

roboz0r:
I changed test projects to dual target `net10.0` and `net481` on Windows only. build.fsx was changed to a build project as this seems to work better when not using .NET SDK 6. 

`paket.lock` rolled forward a lot of dependencies automatically when changing tfm. Not sure if they should all be pinned in the default group?

Numpsy:
fwiw I still want to try looking at #498 again at some point (I keep failing to get time for it, but have an upcoming need for some AOT builds of things at work), and as you need to target .NET 8.0+ to get proper support for NativeAOT I might ask about multi-targetting things at .NET 8.0 and .NET Standard 2.0, rather than only .NET Standard?
(It could be changed later, but maybe it's easier to do it in one go?)

roboz0r:
I can give it a go tomorrow but should be trivial to add the target. Not sure about AOT the big challenge will be F# uses reflection based string printing by default

farlee2121:
@Numpsy I'd be curious to learn more about your AOT requirements sometime.

I've been digging into improved code location, and it looks like our main option there is heavily reflection-based.

Numpsy:
>  Not sure about AOT the big challenge will be F# uses reflection based string printing by default

Yes, the %A stuff is potentially a problem (I do have a cli tool I'm working on at work largely working with an AOT build though)

> I'd be curious to learn more about your AOT requirements sometime.

I started looking at it was back and then stopped because my main task at the time was AOT builds of a CLI app where the integration tests are a separate process and don't need to be AOT themselves, and then I started thinking that an AOT build of just a library built for native use should be tested with a native test framework of some sort, but then I now have a case where it would be useful if I could build an existing Expecto test executable natively.. so, I'll give it another try at least.

> I've been digging into improved code location

'All' I'd be wanting to do is natively compile and run a test exe - no test explorers or code locations - so hopefully that part isn't a problem.

ratsclub:
@farlee2121 It has been a while since I did that change. I remember that it was related to not being able to use Tasks properly when targetting netstandard2.0 alone.

farlee2121:
That's great context. We'll have to be extra mindful to review task behavior. Thanks!

roboz0r:
Is there anything left blocking this one? I think it's ready now with or without the dual targeting to net8.0

farlee2121:
I didn't realize you'd finished the work. I'll try to review it soon

Expecto.Tests.CSharp/Expecto.Tests.CSharp.csproj
```diff
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <AssemblyName>Expecto.Tests.CSharp</AssemblyName>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFrameworks>net10.0;net481</TargetFrameworks>
```
farlee2121: Is there a reason this has a different multi-target than the rest of the packages?

farlee2121: Oh, this is a test project. I missed it because .Tests was in a different order

farlee2121:
@roboz0r The changes mostly made sense to me. I left a few questions.

I've been thinking of migrating fake off of fsx for a while, so thanks for doing that too.

Numpsy:
> I've been thinking of migrating fake off of fsx for a while, so thanks for doing that too.

Does fake-cli still need to be in the tools json if the build script has been converted into a project?

farlee2121:
> > I've been thinking of migrating fake off of fsx for a while, so thanks for doing that too.
> 
> Does fake-cli still need to be in the tools json if the build script has been converted into a project?

Good point. I don't think we need fake-cli anymore.

farlee2121:
I got impatient because this will help with another feature I'm working on.

The only question I didn't resolve is why `System.Runtime.CompilerServices.Unsafe` was included. It didn't appear to be doing anything. @roboz0r What was this package include meant to accomplish?

roboz0r:
Thanks for the review and getting the PR cleaned up. On `System.Runtime.CompilerServices.Unsafe` I don't recall any specific reason I wanted it. Most likely I was chasing warnings about transitive deps and just added it.

PR #532, opened 2026-04-29
Update github actions to fix deprecation warnings

Numpsy:
I just saw these deprecation warnings in the CI builds and thought I'd try updating the actions to fix them:

image


farlee2121:
Thanks for the fix!

